PR #1, opened 2016-03-17
Make build work on OS X

kchodorow:
mktemp takes different arguments on OS (the template is required).

See http://ci.bazel.io/job/rules_closure/BAZEL_VERSION=HEAD,PLATFORM_NAME=darwin-x86_64/1/console.

### DIFF
--- a/third_party/phantomjs/BUILD
+++ b/third_party/phantomjs/BUILD
@@ -11,7 +11,7 @@ genrule(
     cmd = " && ".join([
         "IN=$$(pwd)/$(SRCS)",
         "OUT=$$(pwd)/$@",
-        "TMP=$$(mktemp -d)",
+        "TMP=$$(mktemp -d tmp.XXXXXXXXXX)",
         "cd $$TMP",
         "tar -xjf $$IN",
         "cd phantomjs-*",


third_party/phantomjs/BUILD
```diff
@@ -11,7 +11,7 @@ genrule(
     cmd = " && ".join([
         "IN=$$(pwd)/$(SRCS)",
         "OUT=$$(pwd)/$@",
-        "TMP=$$(mktemp -d)",
+        "TMP=$$(mktemp -d tmp.XXXXXXXXXX)",
```
jart: Out of curiosity, I've seen some bazel code that does it this way: `TMP=$$(mktemp -d $${TMPDIR:-/tmp}/phantomjs.XXXXXXXX)`. Do you know if there's any real difference between the two approaches?

jart:
This will get us closer to compatibility. But we also need to have a `select()` statement that causes the platform-specific binary for PhantomJS to be extracted.

Someone else actually just sent me a PR fixing this, in a private fork of this repository. I can probably integrate that tomorrow.

third_party/phantomjs/BUILD
kchodorow: AFAICT, it's equivalent.

jart:
LGTM

PR #3, opened 2016-03-18
Make PhantomJS fully hermetic

jart:
R=@kchodorow

This commit adds libfontconfig, libfreetype, libexpat, and libpng. We
fetch these prebuilt from the Debian Jessie (Stable) repository.
PhantomJS will now link against them, rather than whatever's on the
system.

Fontconfig has been configured to use the Google Noto
fonts (https://www.google.com/get/noto/). This font family has been
selected because it's cosmopolitan, looks fabulous, and doesn't have
legal encumbrances like DejaVu or msttcorefonts.

I've added a screenshot test that does a binary comparison on the
generated PNG file. With a little bit of luck, this build environment
will be so deterministic that this will actually work.

This commit does not add Darwin support; that comes later.


jart:
Oh my gosh oh my gosh oh my gosh. I *can not believe* that actually passed on Travis on the first try.

jart:
R+= @cydeweys

closure/private/files_equal_test.bzl
```diff
@@ -0,0 +1,61 @@
...
+
+"""Tests that two files contain the same binary data."""
+
+def _impl(ctx):
+  ctx.file_action(
+      output=ctx.outputs.executable,
+      content="\n".join([
+          "#!/bin/bash",
+          "function checksum() {",
+          "  if command -v sha256sum >/dev/null; then",
```
kchodorow: Use `openssl sha256`, it's available basically everywhere.

closure/private/files_equal_test.bzl
```diff
@@ -0,0 +1,61 @@
...
+          "    sha256sum $1 | cut -f 1 -d ' '",
+          "  elif command -v shasum >/dev/null; then",
+          "    cat $1 | shasum -a 256 | cut -f 1 -d ' '",
+          "  else",
+          "    echo no sha256 checksum command found >&2",
+          "    exit 1",
+          "  fi",
+          "}",
+          "SUM1=$(checksum %s)" % ctx.file.golden.short_path,
```
kchodorow: Where is %s set?

closure/private/files_equal_test.bzl
```diff
@@ -0,0 +1,61 @@
...
+          "SUM2=$(checksum %s)" % ctx.file.actual.short_path,
+          "if [[ ${SUM1} != ${SUM2} ]]; then",
+          "  echo FILES DO NOT HAVE EQUAL CONTENTS >&2",
+          "  echo %s ${SUM1} >&2" % ctx.file.golden.short_path,
+          "  echo %s ${SUM2} >&2" % ctx.file.actual.short_path,
+          "  exit 1",
+          "fi",
+      ]),
```
kchodorow: I think you actually need to pass in the strings to replace %s.  Using .format() is probably clearer.

jart: Are you sure you're not mistaken? I'm passing the format arguments in via `%`. I also can't really use `"\n".join([...]).format(golden=..., ...)` here because then it will complain that all the other stuff inside curlies doesn't have a replacement defined.

closure/private/files_equal_test.bzl
jart: Done.

jart:
PTAL

closure/private/files_equal_test.bzl
kchodorow: Sorry, brain fart.

kchodorow:
LGTM

Issue #19, opened 2016-03-24
phantomjs.sh not found


wolfgangmeyers:
I've found the following in the error log when running `bazel test` for `closure_js_test` rules:
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
{test file}: 2: exec: third_party/phantomjs/phantomjs.sh: not found
```

wolfgangmeyers:
I've found that this file is present in the bazel-bin folder:
`bazel-bin/{test file}.runfiles/external/io_bazel_rules_closure/third_party/phantomjs/phantomjs.sh`

wolfgangmeyers:
Looks at first glance like the error originates here (https://github.com/bazelbuild/rules_closure/blob/master/closure/testing/closure_js_test.bzl#L77):
```python
ctx.file_action(
      executable=True,
      output=ctx.outputs.executable,
      content="\n".join([
          "#!/bin/sh",
          "exec third_party/phantomjs/phantomjs.sh \\",
          "  %s \\" % ctx.file._phantomjs_runner.short_path,
          "  %s" % ctx.outputs.js.short_path,
          "",
      ]))
```

wolfgangmeyers:
My guess is that `external/io_bazel_rules_closure` needs to be prepended to the path of phantomjs.sh. I'll test this out and let you know.

wolfgangmeyers:
Well it looks like phantomjs.sh also needs this prepended somewhere:
```
external/io_bazel_rules_closure/third_party/phantomjs/phantomjs.sh: line 38: {test file}.runfiles/third_party/phantomjs/bin/phantomjs: No such file or directory
```

wolfgangmeyers:
hmm, looks like this change broke the tests:
```
bazel-out/host/bin/third_party/phantomjs/phantomjs: line 38: bazel-out/host/bin/third_party/phantomjs/phantomjs.runfiles/external/io_bazel_rules_closure/third_party/phantomjs/bin/phantomjs: No such file or directory
```

jart:
How are you running this?

jart:
How would I reproduce this?

wolfgangmeyers:
I'll get you a minimal project shortly.

jart:
What is the command you are running on the shell though, that produces this error?

wolfgangmeyers:
[...]

PR #20, opened 2016-03-24
Phantomjs path fix

wolfgangmeyers:
phantomjs and phantomjs.sh needed to have `external/io_bazel_rules_closure` prepended to their paths.
This resolves https://github.com/bazelbuild/rules_closure/issues/19


wolfgangmeyers:
Abandoning this fix, since it breaks the tests. We need a solution that works both for client code and for the tests.

PR #29, opened 2016-04-01
Add Closure Library dep to Soy JS library

### LINKED ISSUE #28: closure library required dependency for closure_template_js_library
If compiling without first adding dependency on the closure library:
```
    deps = [
        "@io_bazel_rules_closure//closure/library",
    ],
```
this error is printed when building the template library:
```
JavaScript strict dependency checking failed D:

These namespaces were required by {build target}

	goog.asserts

In the following files:

	bazel-out/local_linux-fastbuild/genfiles/{soy template file}
```

I'll attach a reproduction scenario when I have time - unless it seems sufficient to just document this.

jart:
This fixes #28 and also adds a unit test for Soy.

Reviewer: @wolfgangmeyers


wolfgangmeyers:
LGTM

PR #33, opened 2016-04-05
phantomjs wrapper: fix runfiles tree algorithm

damienmg:
This fix failing test with bazel @HEAD.

Tested: bazel 0.2.1 and bazel @HEAD: bazel clean --expunge && bazel test ...

### DIFF
--- a/third_party/phantomjs/phantomjs.sh
+++ b/third_party/phantomjs/phantomjs.sh
@@ -19,10 +19,13 @@
 #         export FC_DEBUG="1024"
 
 if [[ "${RUNFILES}" == "" ]]; then
-  if [[ "$(basename "$(pwd)")" =~ \.runfiles$ ]]; then
-    RUNFILES="$(pwd)"
+  if [[ "$(pwd)" =~ ^(.*\.runfiles)(/.*)?$ ]]; then
+    RUNFILES="${BASH_REMATCH[1]}"
   else
     RUNFILES="${0}.runfiles"
+  fi
+  if [ -d "${RUNFILES}/io_bazel_rules_closure" ]; then
+    RUNFILES="${RUNFILES}/io_bazel_rules_closure"
   fi
 fi
 


damienmg:
note that ci.bazel.io is currently down (I am restarting it right now so it shouldn't be too long until we can check)

third_party/phantomjs/phantomjs.sh
```diff
@@ -19,11 +19,14 @@
...
 
 if [[ "${RUNFILES}" == "" ]]; then
-  if [[ "$(basename "$(pwd)")" =~ \.runfiles$ ]]; then
-    RUNFILES="$(pwd)"
+  if [[ "$(pwd)" =~ ^(.*\.runfiles)(/.*)?$ ]]; then
+    RUNFILES="${BASH_REMATCH[1]}"
   else
     RUNFILES="${0}.runfiles"
   fi
+  if [ -d "${RUNFILES}/io_bazel_rules_closure" ]; then
```
jart: 1. Since I set the workspace(name), wouldn't this *always* be true? If so, why would we have an if statement?
2. Where will external dependency runfiles be stored? Will it be io_bazel_rules_closure/external/phantomjs? Or external/phantomjs?
3. If it's the prior, then I'm assuming Bazel now transitively fetches WORKSPACE dependencies? If so, should the contents of my repositories.bzl be moved into my WORKSPACE file?

damienmg: 1. Not for released version (0.2.1) who has the bug.
2. for now io_bazel_rules_closure/external/phantomjs/, in the future simply phantomjs/ (with a transition periode where both are correct).
3. No and not planed in the near future. We do not do transitive dependencies fetching for many reason, mostly prevent dependency hell.

jart: I'm so glad there's no transitive dependency fetching and that it isn't planned. It's not worth doing something unless you can do it right.

PR #48, opened 2016-05-03
Expand closure_repositories() API for easy dependency version overrides

hochhaus:
When trying to utilize the version override provided by 17619240422e5b076e784fee0c5da89ca31ffafc I was still doing a good amount of copy/pasting. (However, I could have been doing it wrong. If so, please let me know.)

With this change I am able to specify custom versions of any dependency with very minimal code.

Are you open to an idea like this? Does a better solution exist?

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,28 @@ closure_repositories()
 You are not required to install the Closure Tools or PhantomJS. They will be
 fetched automatically.
 
+### Overriding Dependency Versions
+
+When you call `closure_repositories()` in your `WORKSPACE` file, it causes a
+few dozen external dependencies to be added to your project, e.g. Guava, Guice,
+JSR305, etc. You might need to customize this behavior.
+
+To override the version of any dependency, modify your `WORKSPACE` file to pass
+`omit_<dependency_name>=True` to `closure_repositories()`. Next define your
+custom dependency version. A full list of dependencies is available from
+[repositories.bzl](https://github.com/bazelbuild/rules_closure/tree/master/closure/repositories.bzl).
+For example, to override the version of Guava:
+
+```python
+load("@io_bazel_rules_closure//closure:defs.bzl", "closure_repositories")
+closure_repositories(omit_guava=True)
+
+maven_jar(
+  name = "guava",
+  artifact = "...",
+  sha1 = "...",
+)
+```
 
 ## Examples
 
--- a/closure/repositories.bzl
+++ b/closure/repositories.bzl
@@ -227,33 +227,91 @@ def soyutils_usegoog():
       url = "https://bazel-mirror.storage.googleapis.com/repo1.maven.org/maven2/com/google/template/soy/2016-01-12/soy-2016-01-12-soyutils_usegoog.js",
   )
 
-def closure_repositories():
-  aopalliance()
-  asm()
-  asm_analysis()
-  asm_commons()
-  asm_util()
-  args4j()
-  closure_compiler()
-  closure_library()
-  closure_linter()
-  closure_stylesheets()
-  fonts_noto_hinted_deb()
-  fonts_noto_mono_deb()
-  guava()
-  gson()
-  guice()
-  guice_assistedinject()
-  guice_multibindings()
-  icu4j()
-  jsr305()
-  jsr330_inject()
-  libexpat_amd64_deb()
-  libfontconfig_amd64_deb()
-  libfreetype_amd64_deb()
-  libpng_amd64_deb()
-  phantomjs_linux_x86_64()
-  phantomjs_macosx()
-  python_gflags()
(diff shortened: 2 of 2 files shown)


jart:
Here is an example of how I'm using the design as it's currently written. In a separate project of mine, I put the following in my `WORKSPACE` file.

image

You'll notice that I left out some dependencies from that list. Closure Rules specifies Guava. But I already define Guava in my other project. So I didn't bother importing that dependency from Closure Rules.

What do you think?

jart:
One disadvantage to the alternative design you're proposing, is that if I wanted to specify my own Guava, I would have to define it as a function in my workspace file and then pass that function to `closure_repositories()`.

The advantage to the current design is that it's explicit and straightforward. My workspace file shows all my transitive dependencies. That way I know exactly what external `@foo//` names I can reference. The disadvantage to this design is that if we add a new dependency to Closure Rules, we have to mention in the release notes that users need to add yet another one of those lines to their workspace file.

One thing that might help improve the current design, would be if we coalesced some of these dependencies into groups.

```python
def closure_repositories_asm():
  asm()
  asm_analysis()
  asm_commons()
  asm_util()

def closure_repositories_closure():
  closure_compiler()
  closure_library()
  closure_linter()
  closure_stylesheets()
  soy()
  soyutils_usegoog()

def closure_repositories_guice():
  aopalliance()
  guice()
  guice_assistedinject()
  guice_multibindings()

def closure_repositories_phantomjs():
  fonts_noto_hinted_deb()
  fonts_noto_mono_deb()
  libexpat_amd64_deb()
  libfontconfig_amd64_deb()
  libfreetype_amd64_deb()
  libpng_amd64_deb()
  phantomjs_linux_x86_64()
  phantomjs_macosx()
  
def closure_repositories():
  args4j()
  closure_repositories_asm()
  closure_repositories_closure()
  closure_repositories_guice()
  closure_repositories_phantomjs()
  guava()
  gson()
  icu4j()
  jsr305()
  jsr330_inject()
  python_gflags()
```

That would cut out a lot of the verbosity in my other project's `WORKSPACE` file. It would also allow us to maintain Closure Rules in a way that's less likely to break our users.

hochhaus:
@jart Good point about not requiring extra function definitions being passed to `closure_repositories()`.

What about allowing certain dependencies to be omitted instead? This avoids the need for extra functions, allows for very granular dependency toggling, does not require new dependencies to be announced in the release notes and minimizes the need to load() extra symbols. I updated the change with a specific proposal.

Alternatively, your grouped dependencies proposal also would work from my perspective. Which do you prefer?

README.md
```diff
@@ -68,6 +68,23 @@ closure_repositories()
 You are not required to install the Closure Tools or PhantomJS. They will be
 fetched automatically.
 
+### Overriding Dependency Versions
+
+To override the version of any dependency, modify your `WORKSPACE` file to pass`omit_<dependency_name>=True` to `closure_repositories()`. Next define your
```
jart: 1. Please wrap an column 80.
2. Space after 'pass'
3. It might help if we give the user a little more context. Consider adding a paragraph here that sounds something like: “When you call `closure_repositories()` in your `WORKSPACE` file, it causes a few dozen external dependencies to be added to your project, e.g. Guava, Guice, JSR305, etc. You might need to customize this behavior.”

README.md
```diff
@@ -68,6 +68,23 @@ closure_repositories()
...
+custom dependency version. A full list of dependencies is available from
+[repositories.bzl](https://github.com/bazelbuild/rules_closure/tree/master/closure/repositories.bzl).
```
jart: I really like the fact that you actually linked to the file. That makes life so much easier for the reader.

README.md
```diff
@@ -68,6 +68,23 @@ closure_repositories()
...
+For example, to override the version of closure-compiler:
+
+```python
+load("@io_bazel_rules_closure//closure:defs.bzl", "closure_repositories")
+closure_repositories(omit_closure_compiler=True)
+
+maven_jar(
+  name = "closure_compiler",
```
jart: Suggestion: Use Guava as the example. It's much more likely the user will want to omit that. I'd rather not encourage people to override Closure dependencies. They'd be better off sending us a pull request if they feel the version of Closure we're tracking is too old.

jart:
I must admit, I'm biased against exclusions. I don't think it's idiomatic to Bazel. But I like the way you did this. I think it makes sense. So I'm happy merging this if you address the nits above :)

hochhaus:
@jart Thanks for the detailed feedback. Comments addressed in the updated commit.

If you can think of anyway to make this change more idiomatic for bazel please let me know. Thanks for your flexibility during the design discussion.

jart:
LGTM

Issue #39, opened 2016-04-23
pedantic mode option to make deprecated & unnecessaryCasts errors


hochhaus:
This is more of a question than an issue but I couldn't find a mailing list. What is the best forum for end user discussion about the bazel rules_closure?

When setting pedantic mode, deprecated and unnecessaryCasts are only warnings (not errors).

    JS_PEDANTIC_ARGS = [
        "--jscomp_error=*",
        "--jscomp_warning=deprecated",
        "--jscomp_warning=unnecessaryCasts",
    ]

This can be overcome by passing flags directly to the compiler in the BUILD file, for example:

    defs = [
      "--jscomp_error=deprecated",
      "--jscomp_error=unnecessaryCasts",
    ],

I would expect pedantic mode to make all diagnostic groups errors. Is the best way to override the default pedantic behavior by using defs (as above)? Am I missing a better configuration hook?

jart:
1. I'll create a mailing list later today or tomorrow. I've been meaning to do that.

2. The best option at the moment is to use `defs`. I've been intentionally avoiding creating thin wrappers for command line flags, unless my abstraction can provide additional value.

PR #49, opened 2016-05-04
Add incremental syntax checking and linting

jart:
This is a huge improvement over our old system for doing strict dependency checking. Now we can show better error messages that point to the actual source code. It'll go faster. And we get syntax validation and linting for free! Plus it goes wicked fast. At least one order of a magnitude faster than building a JS binary. So now developers can see errors faster.

Reviewer: @hochhaus 
CC: @concavelenz @MatrixFrog


closure/library/closure_library.BUILD
```diff
@@ -837,23 +837,14 @@ filegroup(
         "closure/goog/labs/testing/assertthat.js",
         "closure/goog/labs/testing/assertthat_test.js",
```
hochhaus: Should assertthat_test.js be removed as well?

closure/repositories.bzl
```diff
@@ -227,33 +227,45 @@ def soyutils_usegoog():
       url = "https://bazel-mirror.storage.googleapis.com/repo1.maven.org/maven2/com/google/template/soy/2016-01-12/soy-2016-01-12-soyutils_usegoog.js",
   )
 
-def closure_repositories():
-  aopalliance()
+def closure_repositories_asm():
```
hochhaus: I think the dependency groups should be reverted per the `omit_<dependency_name>` change.

closure/compiler/closure_js_library.bzl
```diff
@@ -74,8 +83,9 @@ closure_js_library = rule(
         "deps": JS_DEPS_ATTR,
         "exports": JS_DEPS_ATTR,
         "language": attr.string(default=JS_LANGUAGE_DEFAULT),
+        "suppress": attr.string_list(),
```
hochhaus: Does it make sense to add `suppress` support to `closure_js_binary()` for symmetry in the API?

hochhaus: Can you also update `README.md` to document the `suppress` option of ` closure_js_library`?

hochhaus:
@jart This change looks great. I'm excited to use it. As you said a huge improvement over the previous system.

I think it would be helpful to add a brief description of `JsChecker` to `README.md` for new users. As it isn't part of the standard open source closure tools I wouldn't expect it as a new user.

One thing that is unfortunate about this design is that suppressions from the `closure_js_library()` level don't flow to the `closure_js_binary()` level. In the ideal world I think that library suppressions should also propagate to the binary (but only apply to the sources from the corresponding library, not to all of the sources in the binary). To the best of my knowledge this granularity of suppressions isn't possible from the command line today (either with jscomp_off or hide_warnings_for). I don't think you can (easily) resolve this point, but I just wanted to mention it.

Finally, I don't understand the compiler well enough to meaningfully review the `JsChecker*.java` files. However, I did pull your change into my project to test and it seems to work well.

Thanks for implementing this.

closure/repositories.bzl
jart: Whoops didn't mean for that to be in this PR.

closure/library/closure_library.BUILD
jart: Ok this should be completely fixed up now.

jart:
You've matched my thoughts exactly when it comes to how suppressions should propagate upwards. But this CL only really checks genuine bona fide errors one would never want to suppress, and linting. The lint checks aren't enabled on `closure_js_binary()`. So in this PR's current state, it shouldn't be a problem.

However I'd like to iterate on this in follow-up to have a truly integrated error/warning system that is much more friendly than the stock Closure Compiler, and better documented. But that effort is out of scope for this, which is a first start.

As for documenting this new flag, I'd like to do that right before we're ready to cut a new release. Otherwise the user will see documentation for something that isn't available.

jart:
PTAL

hochhaus:
Good point about deferring the documentation update.

I just noticed that some lint errors are caught by `linter.jar` but missed by `JsChecker`. For example:

```
ahochhaus@ahochhaus-pc:~/ll$ cat example/BUILD 
load("@io_bazel_rules_closure//closure:defs.bzl", "closure_js_library")

closure_js_library(
  name = "example_lib",
  srcs = ["example.js"],
  language = "ECMASCRIPT6_STRICT",
  deps = ["@io_bazel_rules_closure//closure/library"],
)
ahochhaus@ahochhaus-pc:~/ll$ cat example/example.js 
goog.module('ll.example');

const dom = goog.require('goog.dom');

(function() {
  goog.dom.getElement('foo');
})();
ahochhaus@ahochhaus-pc:~/ll$ java -jar ~/closure-compiler/linter.jar example/example.js 
example/example.js:3: WARNING - extra require: 'dom'
const dom = goog.require('goog.dom');
            ^

example/example.js:6: WARNING - Reference to fully qualified import name 'goog.dom'. Please use the short name 'dom' instead.
  goog.dom.getElement('foo');
  ^

example/example.js:6: WARNING - missing require: 'goog.dom.getElement'
  goog.dom.getElement('foo');
  ^

0 error(s), 3 warning(s)
ahochhaus@ahochhaus-pc:~/ll$ bazel build //example:example_lib
INFO: Found 1 target...
Target //example:example_lib up-to-date (nothing to build)
INFO: Elapsed time: 0.231s, Critical Path: 0.00s
```

@jart Any idea why these are missing?

I'm trying to look a bit deeper as well.

hochhaus:
@jart I had another look and everything looks good.

I don't have strong feeling on if `suppress` should be added to `closure_js_binary()` for symmetry. Whichever you think is better sounds good to me.

I very much like your future vision of a "truly integrated error/warning system". Let me know if I can help.

I wish I understood the cause of the missing linter.jar errors.

jart:
Good catch! I know what's causing it. I'm working to address it better, and make some more comparisons with linter.jar output.

If you look at [DiagnosticGroups.java](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DiagnosticGroups.java) and [DefaultPassConfig.java](https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/DefaultPassConfig.java) you'll notice that configuring whether or not various warnings actually appear is kind of anarchy. Hopefully I can figure out a way to bring order. Because the whole error/warning system has always been my biggest gripe with the Closure Compiler.

I think this project represents a clean slate for Closure. So my strategy right now is to be as maximally strict and pedantic as possible. Then as we use the Closure Rules, we'll discover which things are too annoying, and slowly back off. I think this is a good strategy. Because if we start off too loose, and tighten up, we'll end up breaking people. Does that sound like a good strategy to you?

hochhaus:
Thanks for fixing this.

I love the idea of being maximally strict. I would suggest that if you need to back off for Google internally, we leave the ability for external users (with much smaller code bases) to opt-into the maximally strict mode. One example of this is Issue #39 where I would like to keep `deprecated` and `unnecessaryCasts` errors by default. I have re-added them using `defs` but if they were backed off even more I would need to audit future changes (to learn about the additional backing off strictness) and then update my rules accordingly.

Possibly it makes sense to have three levels of `pedantic` that are standardized across `closure_rules`? For example:
- 0 = not pedantic
- 1 = as pedantic as is practical internally at Google
- 2 = as pedantic as the tools allow

If we did something like this we could start with pedantic level 2 being the same as pedantic level 1. However, at the point that you need to back off for internal use cases at Google we could modify only level 1 (but leave level 2 maximally strict). That way small external users (such as myself) can set `pedantic = 2` while internal Google projects can set `pedantic = 1`. I think this gives us more flexibility to change the definitions of these levels in the future without breaking user expectations.

What do you think about a design like that?

jart:
Oh it seems the problem is that the warnings you're seeing aren't available in the v20160315 we're tracking. I hope they cut a new release soon. I asked for more info in https://github.com/google/closure-compiler/issues/1770.

hochhaus:
Sorry about that! I should have been more careful in my testing.

On a side note, I like the idea of sync'ing releases with closure-{compiler,library}. I think it is a good idea even if they don't move to a regular release schedule. My reasoning being that then the `WORKSPACE` definition itself makes it clear what version of the closure tools are being used.

```
git_repository(
    name = "io_bazel_rules_closure",
    remote = "https://github.com/bazelbuild/rules_closure.git",
    tag = "v20160315",
)
```

A slow release cycle would be most painful in the beginning while this repo is iterating very quickly. However, I think that problem would (mostly) go away longer term.

This change LGTM

jart:
Yeah once this project matures, we should probably use dated version tags identical to the Closure Compiler, and adopt the exact same release cycle. For now I'm happy with the 0.x.x versions.

Thanks so much for the review!

PR #54, opened 2016-05-11
Improve JS error checking and suppression

jart:
1. Linter warnings may now be suppressed in a fine-grained way. Previously, you could only suppress an entire group of warnings, e.g. `"lintChecks"` which encompasses 20+ diagnostic types. Now you can suppress each individual diagnostic type too, e.g. `"JSC_DUPLICATE_ENUM_VALUE"`, `"JSC_MISSING_JSDOC"`, etc. See `closure/library/BUILD` for an example of this. (I think this is much more readable than the previous system of arbitrary group names. It allows the user to easily search for the compiler code that emits a particular warning.)

2. The Closure Compiler will now *actually tell you* the magic word for suppressing a warning. For example:

    > closure/compiler/test/linting/love.js:51: WARNING **lintChecks JSC_MISSING_JSDOC** - Function must have JSDoc.

    In the example above, either `lintChecks` or `JSC_MISSING_JSDOC` could be added to the `suppress` list in order to hide that warning.

3. Errors are emitted for superfluous suppressions. For example if all your sources have jsdocs, yet you say `suppress = ["JSC_MISSING_JSDOC"]`, the build will fail because it's not necessary to suppress that linter warning. Thus ensuring the long-term maintainability of suppressions.

4. Regression tests are now implemented for strict dependency checking and linting. These actually test failure conditions.

PTAL @hochhaus


hochhaus:
LGTM. Huge set of improvements.

jart:
:smile: So happy you approve.

Issue #52, opened 2016-05-11
//closure/testing:lint, //closure/library/test:lint and //closure/testing/test:noto_fonts_render_as_expected are failing on darwin at HEAD


damienmg:
See https://github.com/bazelbuild/bazel/issues/1241

Probably due to cc94e42

For some reason this commit had not generated a build on ci.bazel.io

jart:
It seems like two things have happened:

1. Screenshot test has slightly different binary output on Darwin. Who knows why. I'll regenerate that on my laptop and verify it looks correct. Should be an easy fix.

2. Python `import` now functions differently than before. I have no idea what changed. I'm not terribly motivated to try and understand it. So I'm just going to send you a PR deleting the old python linter from this project.

jart:
Closure Rules is now green according to Mr. Jenkins. Closing this issue. Over and out.

damienmg:
This is happening again:
http://ci.bazel.io/job/rules_closure/BAZEL_VERSION=latest,PLATFORM_NAME=darwin-x86_64/133/console

:(

jart:
Is there a way I can make that test just not run on darwin?

damienmg:
humm, in Bazel we have a custom set-up to skip some test on some infrastructure (basically if uname contains XXX, exit 0 with a message saying that we skip the test).

You can also deactivate it from ci by putting the "noci" flag (but that will deactivate it from linux too).

jart:
It like the flakiness has been gone for a while. So this seems safe to close. Did you do something special on your end?

damienmg:
Reopening, we have some flakes on the //closure/testing/test:noto_fonts_render_as_expected test, see http://ci.bazel.io/job/rules_closure/BAZEL_VERSION=latest,PLATFORM_NAME=darwin-x86_64/299/console

jart:
Closing due to inactivity.

PR #55, opened 2016-05-11
Remove deprecated Python Closure Linter

jart:
This fixes 50% of #52.

CC @damienmg @hochhaus


hochhaus:
LGTM.

As part of a different change, what do you think of adding a rule to enforce formatting via clang-format? The compiler lint checks are relatively loose on formatting so I believe that closure-library uses clang-format.

jart:
Wow it uses Clang?!

I wish they had chosen something written in Java because the Clang format dependency is 107MB (http://llvm.org/releases/3.7.1/clang+llvm-3.7.1-x86_64-linux-gnu-ubuntu-14.04.tar.xz) and 557MB unzipped. Although the clang-format program itself appears to be 1.5MB. It also doesn't appear to dynamically link against anything weird:

```
master jart@compy:~/code/closure-library$ ldd ./clang+llvm-3.7.1-x86_64-linux-gnu-ubuntu-14.04/bin/clang-format 
        linux-vdso.so.1 =>  (0x00007fff561f2000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007fef910ef000)
        libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007fef90eeb000)
        libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007fef90cc2000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fef90aa4000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fef9088b000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fef90585000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fef90281000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fef9006b000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fef8fca6000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fef912f7000)
```

So it might be possible to ask tar to only extract that single file. Then it should hopefully *just work* in a hermetically sealed linux build environment.

The *problem* is they [don't appear](http://llvm.org/releases/3.7.1/) to have a darwin release, so we might not be able to maintain Mac compatibility, in a manner similar to how Closure Rules currently maintains Mac compatibility for PhantomJS.

If you want to investigate further, be my guest. But I feel we'd probably be much better off is functionality of relative parity could be offered via Java. Possibly by finding someone willing to adapt [google-java-format](https://github.com/google/google-java-format) for JS.

hochhaus:
Thanks for your comments. I agree that clang is a really heavy weight dependency just for formatting.

Your proposed java based solution is good as well.

My only additional feedback is that llvm does make apple releases [3.7.0](http://llvm.org/releases/3.7.0/clang+llvm-3.7.0-x86_64-apple-darwin.tar.xz) and [3.8.0](http://llvm.org/releases/3.8.0/clang+llvm-3.8.0-x86_64-apple-darwin.tar.xz). I'm not sure why it is missing from the 3.7.1 download directory. I just wanted to mention this since I think Mac compatibility would be possible to retain even if we imported clang-format.

I might look into this more as I have time.

PR #58, opened 2016-05-12
Fix closure_js_deps and test it

jart:
- Make `closure_js_deps` actually work.

- Breaking API change: We no longer generate the legacy deps.js file that includes the weird bazel directory names. We only generate the deps-runfiles.js file now.

- Regression tests.


jart:
Sigh. It seems this depswriter.py differs in behavior depending on whether or not sandboxing is enabled. I should just rewrite it in Java.

PR #61, opened 2016-05-14
Change closure_js_binary() main to entry_points

jart:
I got rid of the `main` attribute and introduced `entry_points` and `dependency_mode`. These will make the behavior more similar to the Closure Compiler, except with better error messages and documentation.

PTAL @robfig @hochhaus

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ Please see the test directories within this project for concrete examples of usa
 
 ```python
 load("@io_bazel_rules_closure//closure:defs.bzl", "closure_js_library")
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -235,7 +208,8 @@ closure_js_library(
 ```python
 load("@io_bazel_rules_closure//closure:defs.bzl", "closure_js_binary")
 closure_js_binary(name, deps, main, css, pedantic, debug, language,
```
robfig: do you mean to remove the main attribute from this declaration?

PR #66, opened 2016-05-20
Add deps.js to forward-declare closure-library types

hochhaus:
https://github.com/google/closure-library/wiki/Frequently-Asked-Questions#when-i-compile-with-type-checking-on-i-get-warnings-about-bad-type-annotation-unknown-type-from-closure-library-files

### DIFF
--- a/closure/testing/closure_js_test.bzl
+++ b/closure/testing/closure_js_test.bzl
@@ -24,6 +24,7 @@
 
 load("//closure/private:defs.bzl",
      "CLOSURE_LIBRARY_BASE_ATTR",
+     "CLOSURE_LIBRARY_DEPS_ATTR",
      "JS_DEPS_ATTR",
      "JS_HIDE_WARNING_ARGS",
      "JS_LANGUAGE_DEFAULT",
@@ -104,6 +105,7 @@ _closure_js_test = rule(
         "pedantic": attr.bool(default=False),
         "defs": attr.string_list(),
         "_closure_library_base": CLOSURE_LIBRARY_BASE_ATTR,
+        "_closure_library_deps": CLOSURE_LIBRARY_DEPS_ATTR,
         "_compiler": attr.label(
             default=Label("//closure/compiler"),
             executable=True),
--- a/closure/compiler/closure_js_binary.bzl
+++ b/closure/compiler/closure_js_binary.bzl
@@ -19,6 +19,7 @@
 
 load("//closure/private:defs.bzl",
      "CLOSURE_LIBRARY_BASE_ATTR",
+     "CLOSURE_LIBRARY_DEPS_ATTR",
      "JS_DEPS_ATTR",
      "JS_LANGUAGE_DEFAULT",
      "JS_PEDANTIC_ARGS",
@@ -160,6 +161,7 @@ closure_js_binary = rule(
             default=Label("//closure/compiler"),
             executable=True),
         "_closure_library_base": CLOSURE_LIBRARY_BASE_ATTR,
+        "_closure_library_deps": CLOSURE_LIBRARY_DEPS_ATTR,
     },
     outputs={
         "out": "%{name}.js",
--- a/closure/compiler/closure_js_deps.bzl
+++ b/closure/compiler/closure_js_deps.bzl
@@ -16,7 +16,9 @@
 
 """Build definitions for JavaScript dependency files."""
 
-load("//closure/private:defs.bzl", "CLOSURE_LIBRARY_BASE_ATTR")
+load("//closure/private:defs.bzl",
+     "CLOSURE_LIBRARY_BASE_ATTR",
+     "CLOSURE_LIBRARY_DEPS_ATTR")
 
 def _impl(ctx):
   # XXX: Other files in same directory will get schlepped in w/o sandboxing.
@@ -75,6 +77,7 @@ closure_js_deps = rule(
             allow_files=False,
             providers=["transitive_js_srcs"]),
         "_closure_library_base": CLOSURE_LIBRARY_BASE_ATTR,
+        "_closure_library_deps": CLOSURE_LIBRARY_DEPS_ATTR,
         "_depswriter": attr.label(
             default=Label("@closure_library//:depswriter"),
             executable=True),
--- a/closure/compiler/closure_js_library.bzl
+++ b/closure/compiler/closure_js_library.bzl
@@ -18,6 +18,7 @@
 
 load("//closure/private:defs.bzl",
      "CLOSURE_LIBRARY_BASE_ATTR",
+     "CLOSURE_LIBRARY_DEPS_ATTR",
      "JS_LANGUAGE_DEFAULT",
      "JS_DEPS_ATTR",
      "JS_FILE_TYPE",
@@ -106,6 +107,7 @@ closure_js_library = rule(
             default=Label("//java/com/google/javascript/jscomp:jschecker"),
             executable=True),
         "_closure_library_base": CLOSURE_LIBRARY_BASE_ATTR,
+        "_closure_library_deps": CLOSURE_LIBRARY_DEPS_ATTR,
     },
     outputs={
         "provided": "%{name}-provided.txt",
--- a/closure/library/closure_library.BUILD
+++ b/closure/library/closure_library.BUILD
@@ -14,12 +14,13 @@
 
 package(default_visibility = ["//visibility:public"])
 
-exports_files(["closure/goog/base.js"])
+exports_files(["closure/goog/base.js", "closure/goog/deps.js"])
 
 filegroup(
     name = "js_files",
     srcs = [
         "closure/goog/base.js",
+        "closure/goog/deps.js",
         ":js_library_files",
         ":js_testing_files",
     ],
--- a/closure/private/defs.bzl
+++ b/closure/private/defs.bzl
@@ -64,6 +64,11 @@ CLOSURE_LIBRARY_BASE_ATTR = attr.label(
     allow_files=True,
     single_file=True)
 
+CLOSURE_LIBRARY_DEPS_ATTR = attr.label(
+    default=Label("@closure_library//:closure/goog/deps.js"),
+    allow_files=True,
+    single_file=True)
+
 def collect_js_srcs(ctx):
   srcs = set(order="compile")
   externs = set(order="compile")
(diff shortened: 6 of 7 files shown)


bazel-io:

closure/testing/closure_js_test.bzl
```diff
@@ -104,6 +105,7 @@ _closure_js_test = rule(
         "pedantic": attr.bool(default=False),
         "defs": attr.string_list(),
         "_closure_library_base": CLOSURE_LIBRARY_BASE_ATTR,
+        "_closure_library_deps": CLOSURE_LIBRARY_DEPS_ATTR,
```
jart: There's no type checking in tests.

jart:
Thanks for this PR. This spots a very important issue I didn't consider.

My concern is I'm not sure if this solves the problem. Wouldn't the files referenced in `deps.js` need to be passed as inputs to the Closure Compiler? For example if I say:

```javascript
/**
 * @param {goog.events.EventType} x
 * @return {goog.events.EventType}
 * @export
 */
function foo(x) {
  return x + "lol";
}
```

```python
closure_js_library(
    name = "foo",
    srcs = ["foo.js"],
)

closure_js_binary(
    name = "foo_bin",
    deps = [":foo"],
)
```

Then only `deps.js`, `base.js`, and `foo.js` get added to the `--js` flags. So would the Closure Compiler then try to load `external/closure_library/closure/goog/events/eventtype.js` only to get a Bazel sandbox error? If that's the case, we would need to find some way to enhance JsChecker to do strict dependency checking on the JSDoc annotations.

This is a problem that needs to be solved in a universal way too. `deps.js` may be available for the Closure Library, but it's not available by default for user-defined libraries. So maybe we would have to find some way to generate a mini `deps.js` file for each `closure_js_library`.

hochhaus:
@jart Thanks.

> Wouldn't the files referenced in deps.js need to be passed as inputs to the Closure Compiler?

I don't think so. See example below for my justification.

> So would the Closure Compiler then try to load external/closure_library/closure/goog/events/eventtype.js only to get a Bazel sandbox error?

I don't think that closure-compiler will read that file at all unless it is (transitively) goog.required. I think my example below covers this case.

To test I grabbed the new `v20160517` {compiler,library} releases and pulled just `base.js`, `deps.js` and `object.js` from the library release into a new directory. (I selected `goog.object` over `goog.events.EventType` as `goog.object` does not have any dependencies.)

```
ahochhaus@ahochhaus-pc:~/test$ ls
base.js  compiler.jar  test.js
ahochhaus@ahochhaus-pc:~/test$ cat test.js 
/**
 * @param {goog.object} x
 * @return {goog.object}
 * @export
 */
function foo(x) {
    return x + "lol";
}
ahochhaus@ahochhaus-pc:~/test$ java -jar compiler.jar \
    --compilation_level=ADVANCED \
    --warning_level=VERBOSE \
    --jscomp_error=* \
    --hide_warnings_for=base.js \
    --dependency_mode=LOOSE \
    --generate_exports \
    --new_type_inf \
    --js base.js \
    --js test.js
test.js:2: ERROR - Type annotation references non-existent type goog.object.
 * @param {goog.object} x
           ^

test.js:3: ERROR - Type annotation references non-existent type goog.object.
 * @return {goog.object}
            ^

2 error(s), 0 warning(s)
ahochhaus@ahochhaus-pc:~/test$ cp ../closure-library-20160517/closure/goog/deps.js .
ahochhaus@ahochhaus-pc:~/test$ java -jar compiler.jar \
    --compilation_level=ADVANCED \
    --warning_level=VERBOSE \
    --jscomp_error=* \
    --hide_warnings_for=base.js \
    --dependency_mode=LOOSE \
    --generate_exports \
    --new_type_inf \
    --js base.js \
    --js deps.js \
    --js test.js
function a(e){return e+"lol"}var b=["foo"],c=this;b[0]in c||!c.execScript||c.execScript("var "+b[0]);for(var d;b.length&&(d=b.shift());)b.length||void 0===a?c[d]?c=c[d]:c=c[d]={}:c[d]=a;
ahochhaus@ahochhaus-pc:~/test$ echo -e "goog.require('goog.object');\n$(cat test.js)" > test.js
ahochhaus@ahochhaus-pc:~/test$ cp ../closure-library-20160517/closure/goog/object/object.js .
ahochhaus@ahochhaus-pc:~/test$ java -jar compiler.jar \
    --compilation_level=ADVANCED \
    --warning_level=VERBOSE \
    --jscomp_error=* \
    --hide_warnings_for=base.js \
    --dependency_mode=LOOSE \
    --generate_exports \
    --new_type_inf \
    --js base.js \
    --js test.js
test.js:1: ERROR - required "goog.object" namespace never provided
goog.require('goog.object');
^

1 error(s), 0 warning(s)
ahochhaus@ahochhaus-pc:~/test$ java -jar compiler.jar \
    --compilation_level=ADVANCED \
    --warning_level=VERBOSE \
    --jscomp_error=* \
    --hide_warnings_for=base.js \
    --dependency_mode=LOOSE \
    --generate_exports \
    --new_type_inf \
    --js base.js \
    --js deps.js \
    --js test.js
test.js:1: ERROR - required "goog.object" namespace never provided
goog.require('goog.object');
^

1 error(s), 0 warning(s)
ahochhaus@ahochhaus-pc:~/test$ java -jar compiler.jar \
    --compilation_level=ADVANCED \
    --warning_level=VERBOSE \
    --jscomp_error=* \
    --hide_warnings_for=base.js \
    --hide_warnings_for=object.js \
    --dependency_mode=LOOSE \
    --generate_exports \
    --new_type_inf \
    --js base.js \
    --js object.js \
    --js test.js
function a(e){return e+"lol"}var b=["foo"],c=this;b[0]in c||!c.execScript||c.execScript("var "+b[0]);for(var d;b.length&&(d=b.shift());)b.length||void 0===a?c[d]?c=c[d]:c=c[d]={}:c[d]=a;
ahochhaus@ahochhaus-pc:~/test$ java -jar compiler.jar \
    --compilation_level=ADVANCED \
    --warning_level=VERBOSE \
    --jscomp_error=* \
    --hide_warnings_for=base.js \
    --hide_warnings_for=object.js \
    --dependency_mode=LOOSE \
    --generate_exports \
    --new_type_inf \
    --js base.js \
    --js deps.js \
    --js object.js \
    --js test.js
function a(e){return e+"lol"}var b=["foo"],c=this;b[0]in c||!c.execScript||c.execScript("var "+b[0]);for(var d;b.length&&(d=b.shift());)b.length||void 0===a?c[d]?c=c[d]:c=c[d]={}:c[d]=a;
```

I wish that the compiler would use the full type information (of, for example, goog.object) to make adding the string "lol" a compile error. However, thus far, I have not found the correct set of command line flags to ensure this behavior.

Do I misunderstanding something?

closure/testing/closure_js_test.bzl
hochhaus: True, however "bazel test closure/..." fails without this line as `collect_js_srcs()` is used by `closure_js_test()`. If you have concerns with this I can modify `collect_js_srcs()` to not include `deps.js` for the testing case however I'm not sure that it is worth adding a (small) special case just to omit `deps.js` from tests.

100% your call. I'll gladly make the change if you like.

hochhaus:
Rebased. No rush on review.

PTAL

jart:
Ok this PR is far from the ideal solution. But it gets us closer to where we want to be. So LGTM.

Jenkins: test this please

jart:
> I wish that the compiler would use the full type information (of, for example, goog.object) to make adding the string "lol" a compile error. However, thus far, I have not found the correct set of command line flags to ensure this behavior.

Yeah that's what I feared. I'm going to try poking around with this issue to see if I can understand it better.

jart:
Ok I thought more about this. 

> Only goog.require() what your javascript code directly needs and uses to run. ([Source](https://github.com/google/closure-library/wiki/Frequently-Asked-Questions#when-i-compile-with-type-checking-on-i-get-warnings-about-bad-type-annotation-unknown-type-from-closure-library-files))

I don't think I agree with this advice. I don't understand why weakening the type checker would be a good idea. The Closure Compiler is great at optimizing away dead code in ADVANCED mode. So it seems like the only solid argument here, is that raw sources mode will go slower. I don't find that argument compelling.

I wrote some code to verify behavior. It turns out you have to use `goog.require` in order for type checking to work. *Even if the file declaring that type was passed as a `--js` argument to the compiler before that file.*

The solution proposed by the wiki (passing the deps file to forward declare the type) basically seems to just suppress the non-existent type error and type checking for that particular type becomes silently disabled. This goes against one of the stated goals of this project, which is *strictness*.

If the Closure Compiler doesn't understand what a type is, then I want to know about it. Then I can choose to either `goog.require` the namespace, or explicitly `goog.forwardDeclare` the type.

Looking at `goog.forwardDeclare`, it doesn't seem to me like it was designed for this use case. It seems to be something you want to use when you want a type documented, but it creates cycles. Or in base.js it's used for the browser types it references, so you have the freedom to not have to pass the BROWSER externs definitions to the compiler.

So here's what I propose:

1. I'll roll this change back.
2. I'll patch Emacs js2-closure so it automatically inserts goog.require statements for types only referenced in JSDoc.

Would that be acceptable on your end? Would you be able to update your code to use `goog.forwardDeclare` or `goog.require`?

hochhaus:
Thanks for giving this extra thought.

> I don't understand why weakening the type checker would be a good idea.

Agreed. This "sloppy" type checking has been a source of frustration for me in the past.

> So it seems like the only solid argument here is that using goog.require exclusively for type definitions will make raw sources mode slower.

Well I think it makes the compilation step for ADVANCED mode marginally slower as well (since the goog.required file must be parsed and the dead code eliminated). I agree that this cost is clearly worth it though.

> It turns out you have to use goog.require in order for type checking to work. Even if the file declaring that type was passed as a --js argument to the compiler before that file.

This matched my understanding.

> I want to start from a place of maximal strictness, and work my way back.

I thoroughly support this design goal. I think you are looking at the problem in the correct manner.

> Would you be able to update your code to use goog.forwardDeclare or goog.require?

Yes, I will happily update my code to use goog.require(). My concern is that will generate an "extraRequire" error. What I really want is to get the extraRequire only when the type is not used in ccode or the JS docs.

Do you think it would be possible to coordinate such a change with the compiler folks?

jart:
Actually it turns out that the best practice these days is to use
goog.forwardDeclare(). I've updated js2-closure to do this. But I've held
off on updating this in closure rules for the time being.

hochhaus:
Sounds good to me. Using `goog.forwardDeclare()` is at least as good as the deps.js solution. Feel free to update rules_closure as you like.

PR #68, opened 2016-05-23
Add Closure-style JavaScript protocol buffer utilities.

hochhaus:
Is adding (closure-style) protocol buffer support to this library something that you are willing to consider?

My reasons for thinking this rule fits well here are that the proto3 JS output can be tightly coupled with `closure_js_library()` making protocol buffers trivial to use from JavaScript. If a `js_proto_library()` rule were added to the upstream protobuf repository it would likely not be integrated as well into the rules_closure ecosystem.

If you prefer this as a separate project it is easy for me to split it to out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 # Closure Rules for Bazel (αlpha) [![Build Status](https://travis-ci.org/bazelbuild/rules_closure.svg?branch=master)](https://travis-ci.org/bazelbuild/rules_closure)
 
-JavaScript | Templating | Stylesheets
---- | --- | ---
-[closure_js_library](#closure_js_library) | [closure_template_js_library](#closure_template_js_library) | [closure_css_library](#closure_css_library)
-[closure_js_binary](#closure_js_binary) | [closure_template_java_library](#closure_template_java_library) | [closure_css_binary](#closure_css_binary)
-[closure_js_deps](#closure_js_deps) | [closure_template_py_library](#closure_template_py_library) |
-[closure_js_test](#closure_js_test) | |
+JavaScript | Templating | Stylesheets | Protocol Buffers
+--- | --- | --- | ---
+[closure_js_library](#closure_js_library) | [closure_template_js_library](#closure_template_js_library) | [closure_css_library](#closure_css_library) | [closure_proto_js_library](#closure_proto_js_library)
+[closure_js_binary](#closure_js_binary) | [closure_template_java_library](#closure_template_java_library) | [closure_css_binary](#closure_css_binary) |
+[closure_js_deps](#closure_js_deps) | [closure_template_py_library](#closure_template_py_library) | |
+[closure_js_test](#closure_js_test) | | |
 
 ## Overview
 
@@ -49,6 +49,9 @@ Closure Rules bundles the following tools and makes them "just work."
   unit tests in a command line environment.
 - [Bazel][bazel]: The build system Google uses to manage a repository with
   petabytes of code.
+- [Protocol Buffers][protobuf]: Google's language-neutral, platform-neutral,
+  extensible mechanism for serializing structured data. This is used instead of
+  untyped JSON.
 
 The Closure Tools were released to the public in 2009, but had previously been
 quite difficult to configure. They were originally designed to be used with
@@ -117,6 +120,7 @@ Please see the test directories within this project for concrete examples of usa
 - [//closure/library/test](https://github.com/bazelbuild/rules_closure/tree/master/closure/library/test)
 - [//closure/templates/test](https://github.com/bazelbuild/rules_closure/tree/master/closure/templates/test)
 - [//closure/stylesheets/test](https://github.com/bazelbuild/rules_closure/tree/master/closure/stylesheets/test)
+- [//protobuf/test](https://github.com/bazelbuild/rules_closure/tree/master/protobuf/test)
 
 
 # Reference
@@ -748,6 +752,47 @@ The documentation on using Closure Stylesheets can be found
   `bazel run @io_bazel_rules_closure//closure/stylesheets -- --help`
 
 
+## closure\_proto\_js\_library
+
+```python
+load("@io_bazel_rules_closure//closure:defs.bzl", "closure_proto_js_library")
+closure_proto_js_library(name, srcs, add_require_for_enums, binary,
+                         import_style)
+```
+
+Defines a set of Protocol Buffer files.
+
+Documentation: [Protocol Buffers][protobuf] [JS][protobuf-js]
+[Generator Options](https://github.com/google/protobuf/blob/master/src/google/protobuf/compiler/js/js_generator.h).
+
+#### Implicit Output Targets
+
+- *name*.js: A generated protocol buffer JavaScript library.
+
+### Arguments
+
+- **name:** ([Name][name]; required) A unique name for this rule. Convention
+  states that such rules be named `foo_proto`.
+
+- **srcs:** (List of [labels][labels]; required) A list of `.proto` source
+  files that represent this library.
+
+- **add_require_for_enums:** (Boolean; optional; default is `False`) Add a
+  `goog.require()` call for each enum type used. If false, a forward
+  declaration with `goog.forwardDeclare` is produced instead.
+
+- **binary:** (Boolean; optional; default is `True`) Enable binary-format
+  support.
+
+- **import_style:** (String; optional; default is `IMPORT_CLOSURE`) Specifies
+  the type of imports that should be used. Valid values are:
+
+  - `IMPORT_CLOSURE`    // goog.require()
+  - `IMPORT_COMMONJS`   // require()
+  - `IMPORT_BROWSER`    // no import statements
+  - `IMPORT_ES6`        // import { member } from ''
+
+
 
 [ClosureCodingConvention]: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/ClosureCodingConvention.java
 [GoogleCodingConvention]: https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/GoogleCodingConvention.java
@@ -778,4 +823,6 @@ The documentation on using Closure Stylesheets can be found
 [managing-dependencies]: https://github.com/google/closure-compiler/wiki/Managing-Dependencies
 [phantomjs-bug]: https://github.com/ariya/phantomjs/issues/14028
 [phantomjs]: http://phantomjs.org/
+[protobuf]: https://github.com/google/protobuf
+[protobuf-js]: https://github.com/google/protobuf/tree/master/js
 [verbose]: https://github.com/google/closure-library/blob/master/closure/goog/html/safehtml.js
--- a/protobuf/BUILD
+++ b/protobuf/BUILD
@@ -1,0 +1,27 @@
+# Copyright 2016 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+load("//closure/compiler:closure_js_library.bzl", "closure_js_library")
+
+closure_js_library(
+    name = "jspb",
+    srcs = ["@protobuf_js//:proto_js_library_files"],
+    suppress = [
+        "JSC_EXTRA_REQUIRE_WARNING",
+        "JSC_OPTIONAL_PARAM_NOT_MARKED_OPTIONAL",
+    ],
+    deps = ["//closure/library"],
+)
--- a/protobuf/closure_proto_js_library.bzl
+++ b/protobuf/closure_proto_js_library.bzl
@@ -1,0 +1,62 @@
+# -*- mode: python; -*-
+#
+# Copyright 2016 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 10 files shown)


bazel-io:

jart:
It is ABSOLUTELY something I'm willing to consider! If we can get protos integrated properly and elegantly, so it's easy to have fully type safe data structures in JS and SOY... then it's game over. That's it. Greatest open source frontend development framework imaginable. I'll review shortly.

@damienmg @kchodorow Should we start a new bazelbuild/rules_proto project?

README.md
```diff
@@ -1,11 +1,11 @@
...
 
-JavaScript | Templating | Stylesheets
---- | --- | ---
-[closure_js_library](#closure_js_library) | [closure_template_js_library](#closure_template_js_library) | [closure_css_library](#closure_css_library)
-[closure_js_binary](#closure_js_binary) | [closure_template_java_library](#closure_template_java_library) | [closure_css_binary](#closure_css_binary)
-[closure_js_deps](#closure_js_deps) | [closure_template_py_library](#closure_template_py_library) |
-[closure_js_test](#closure_js_test) | |
+JavaScript | Templating | Stylesheets | Protocol Buffers
+--- | --- | --- | ---
+[closure_js_library](#closure_js_library) | [closure_template_js_library](#closure_template_js_library) | [closure_css_library](#closure_css_library) | [closure_proto_js_library](#closure_proto_js_library)
```
jart: FYI: I've been meaning to test a theory I have that could potentially allow me to integrate `closure_template_{js,java,py}_library` into a single `closure_template_library` statement. If I'm able to pull that off, the design of this will be changed.

hochhaus:
@jart Awesome, thanks!

Getting type safe data passed all the way to the templates sounds like my ideal setup!

No rush on review.

README.md
```diff
@@ -49,6 +49,8 @@ Closure Rules bundles the following tools and makes them "just work."
   unit tests in a command line environment.
 - [Bazel][bazel]: The build system Google uses to manage a repository with
   petabytes of code.
+- [Protocol Buffers][protobuf]: Google's language-neutral, platform-neutral,
+  extensible mechanism for serializing structured data.
```
jart: Consider adding: This is used instead of untyped JSON.

closure/protobuf/protobuf_js.BUILD
```diff
@@ -0,0 +1,16 @@
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])  # BSD
+
+filegroup(
+    name = "proto_js_library_files",
+    srcs = glob(
+        ["js/*.js", "js/binary/*.js"],
```
jart: Can we change this from a blacklist to a whitelist?

closure/protobuf/BUILD
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+load("//closure/compiler:closure_js_library.bzl", "closure_js_library")
+
+closure_js_library(
+    name = "protobuf",
```
jart: jspb might be a more appropriate name for this rule.

jart: Also it might be worth moving //closure/protobuf to //protobuf which is probably the directory we'll want to use when we make rules_protobuf

third_party/protobuf/BUILD
```diff
@@ -0,0 +1,31 @@
...
+        "//conditions:default": ["@protoc_linux_x86_64//file"],
+    }),
+    outs = ["protoc"],
+    cmd = " && ".join([
+        "IN=$$(pwd)/$(SRCS)",
+        "OUT=$$(pwd)/$@",
+        "TMP=$$(mktemp -d $${TMPDIR:-/tmp}/genrule.XXXXXXXXXX)",
+        "cd $$TMP",
+    ]) + select({
+        ":darwin": " && unzip -q $$IN && ",
```
jart: 1. Is the select necessary here, considering the two statements are identical?
2. Is it possible to pass a parameter to unzip that says, only extract this specific file?

README.md
hochhaus: Sounds good. Collapsing the rules would make the design cleaner.

hochhaus:
:+1: for a bazelbuild/rules_protobuf project. This would give a nice location for golang/protobuf bazel support!

hochhaus:
I rebased my work and addressed the first round of comments.

PTAL

protobuf/closure_proto_js_library.bzl
```diff
@@ -0,0 +1,68 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Utilities for building JavaScript Protocol Buffers.
+"""
+
+load("//closure/compiler:closure_js_library.bzl", "closure_js_library")
+
+def closure_proto_js_library(
```
jart: [Optional] If you want to go for the gold, try defining this as a Skylark `rule`. If the rule returns `transitive_js_srcs`, `transitive_js_externs`, `js_language`, etc. in its `struct()` then it (sort of like duck typing) will be able to behave just like a `closure_js_library` rule from the perspective of parent rules.

protobuf/BUILD
```diff
@@ -0,0 +1,27 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+load("//closure/compiler:closure_js_library.bzl", "closure_js_library")
+
+closure_js_library(
+    name = "jspb",
+    srcs = ["@protobuf_js//:proto_js_library_files"],
+    suppress = [
+        "JSC_EXTRA_REQUIRE_WARNING",
```
jart: [Optional] Consider filing upstream bugs for these.

protobuf/test/example.js
```diff
@@ -0,0 +1,45 @@
...
+
+
+
+/**
+ * Example page.
+ * @param {string} field Value to set for "field".
+ * @constructor
+ * @final
+ */
+io.bazel.rules.closure.protobuf.Example = function(field) {
```
jart: Nice job following Google JavaScript style in this file. I suspect you did a lot of JS dev during your tenure at Google.

jart:
I don't see any obvious problems. I'll take a closer look when it's not super late and probably LGTM tomorrow or the day after.

protobuf/closure_proto_js_library.bzl
hochhaus: Thanks. I'll look into doing this soon. Please hold off on further review.

protobuf/BUILD
hochhaus: https://github.com/google/protobuf/issues/1602

protobuf/test/example.js
hochhaus: Strangely enough, I did very little. I don't think I even had JS readability. After closure was open sourced I started using it heavily and learned a lot from reading the library and style guide as well as getting feedback from gjslint. Thanks.

protobuf/closure_proto_js_library.bzl
hochhaus: I looked briefly at this. Originally I had selected implementing `closure_template_js_library()` as a macro because I was composing `native.genrule()` and `closure_js_library()`. This is very similar to how `closure_template_js_library()` first generates the SOY file and then the JS library.

Is it practical to both steps in a single rule?

jart: In this case, absolutely. The `closure_template_*_library` macro is also going to be rewritten as a Skylark rule in the future. The only reason it's a macro, is because I just copied the legacy code outside of g3 in the interest of saving time. But still, optional. If you don't do it, I probably will in the future. But you would benefit from learning more about Skylark if you haven't already. It's pretty cool.

README.md
```diff
@@ -748,6 +752,54 @@ The documentation on using Closure Stylesheets can be found
...
+### Arguments
+
+- **name:** ([Name][name]; required) A unique name for this rule. Convention
+  states that such rules be named `foo_proto`.
+
+- **srcs:** (List of [labels][labels]; required) A list of `.proto` source
+  files that represent this library.
+
+- **add_require_for_enums:** (Boolean; optional; default is `False`) Add a
+  `goog.requires()` call for each enum type used. If false, a forward
```
jart: `goog.require`

protobuf/closure_proto_js_library.bzl
hochhaus: Thanks. I know next to nothing about bazel and I want to learn. Since I now know that is it possible I'm looking forward to converting this to a rule. :)

protobuf/test/example.js
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Message
+   * @private {!proto.io.bazel.rules.closure.protobuf.Message}
+   * @const
+   */
+  this.message_ = new proto.io.bazel.rules.closure.protobuf.Message();
+  this.message_.setField(field);
```
jart: [Optional] [Nit]: `setField` looks like a generic method rather than something that sets the field named `field`. Perhaps rename `field` to `foo`?

README.md
```diff
@@ -748,6 +752,54 @@ The documentation on using Closure Stylesheets can be found
...
+
+- **namespace_prefix:** (String; optional; default is `None`) Specify this
+  to prefix the generated namespace.
+
+- **error_on_name_conflict:** (Boolean; optional; default is `True`) Errors
+  if there are two types that would generate the same output file.
+
+- **binary:** (Boolean; optional; default is `True`) Enable binary-format support.
+
+- **import_style:** (String; optional; default is `None`) Specifies the type
```
jart: Shouldn't default be `"IMPORT_CLOSURE"`?

Also side note: I'm torn about whether or not it would be worth it to automatically add the `IMPORT_` suffix for the user. I'll support you if you want to do that. But either way is fine. Might be better to follow same convention as the tool itself.

README.md
```diff
@@ -748,6 +752,54 @@ The documentation on using Closure Stylesheets can be found
...
+  files that represent this library.
+
+- **add_require_for_enums:** (Boolean; optional; default is `False`) Add a
+  `goog.requires()` call for each enum type used. If false, a forward
+  declaration with `goog.forwardDeclare` is produced instead.
+
+- **namespace_prefix:** (String; optional; default is `None`) Specify this
+  to prefix the generated namespace.
+
+- **error_on_name_conflict:** (Boolean; optional; default is `True`) Errors
```
jart: Why *wouldn't* we want an error to happen in this situation?

protobuf/closure_proto_js_library.bzl
jart: Oh and one other note: if you do convert these to skylark rules, it would probably be better in a follow-up PR.

README.md
```diff
@@ -748,6 +752,54 @@ The documentation on using Closure Stylesheets can be found
...
+  states that such rules be named `foo_proto`.
+
+- **srcs:** (List of [labels][labels]; required) A list of `.proto` source
+  files that represent this library.
+
+- **add_require_for_enums:** (Boolean; optional; default is `False`) Add a
+  `goog.requires()` call for each enum type used. If false, a forward
+  declaration with `goog.forwardDeclare` is produced instead.
+
+- **namespace_prefix:** (String; optional; default is `None`) Specify this
```
jart: OOC when would this be useful? Like if you downloaded a proto file from an external source and wanted to compile it with a different package name? Seems like an unusual feature. I would be perfectly fine leaving weird features out in the interest of keeping the documentation simpler, until we know for sure we need them.

README.md
```diff
@@ -748,6 +752,54 @@ The documentation on using Closure Stylesheets can be found
...
+
+### Arguments
+
+- **name:** ([Name][name]; required) A unique name for this rule. Convention
+  states that such rules be named `foo_proto`.
+
+- **srcs:** (List of [labels][labels]; required) A list of `.proto` source
+  files that represent this library.
+
+- **add_require_for_enums:** (Boolean; optional; default is `False`) Add a
```
jart: Note: It seems like this flag is related to the PR you sent me a little bit ago, re: whether or not a `goog.require` is necessary for `@param` types. I would like for Closure Rules to be able to abstract this blemish in the future. We should probably ask the compiler / library guys if it would be possible for us to decide on *one* correct way to do it, and then enforce it in JsChecker. Maybe we should have a tracking issue for this.

protobuf/closure_proto_js_library.bzl
hochhaus: Sounds good. Once this is merged I'll look into converting the macro to a rule.

README.md
hochhaus: Fixed.

protobuf/test/example.js
hochhaus: Fixed.

README.md
hochhaus: Changed documented default to "IMPORT_CLOSURE". In terms of the implementation which is more idiomatic bazel (None which defers to the plugins default of IMPORT_CLOSURE or an explicit IMPORT_CLOSURE which is passed to the plugin?

I'm happy naming the values with or without the "IMPORT_" prefix. My initial thought was to use the same values as the underlying tool.

README.md
hochhaus: To me it seems like we would always want this behavior. I was surprised it is False by default in protoc.

Can you check if any compelling use cases exist internally and if not I'll remove the flag?

README.md
hochhaus: I agree. I can't think of compelling use cases for namespace_prefix or error_on_name_conflict.

Can you do a quick search internally to see if either have interesting use cases? If not, I'll remove them both.

README.md
hochhaus: Unless I mis-understand (which is very possible) when a type is forward declared (but not required elsewhere) the compiler's level of type checking is binary (either the type exists or it does not). The compiler knows nothing else about the implementation of the type (eg: data members, functions, etc). However, as the compiler works on all sources at a time I think the most common (only?) case this can happen in is code that takes a type as a function parameter but does not need to require it that type. I think this happens most frequently (only?) for code which is not called. If the code were called some other file would have needed to instantiate said type -- hence creating an explicit dependency and enabling the compilers full type checking.

Do you know of more complicated cases?

I agree that a lot more thought and discussion is needed here and an issue tracker item would be good.

jart: I filed a bug for me to look into this: https://github.com/bazelbuild/rules_closure/issues/74

README.md
jart: The only thing in the entire google codebase that references this flag is the protobuf sauce itself. So you can remove this.

jart:
LGTM

Please remove the flag I mentioned and rebase on top of master.

hochhaus:
Thanks.

I removed `namespace_prefix` and hard coded `error_on_name_conflict` to always be on. Finally I rebased.

PTAL

jart:
Hey Jenkins: test this please

PR #71, opened 2016-05-24
Update {compiler,library} to v20160517.

hochhaus:
This change would also need http://dl.google.com/closure-compiler/compiler-20160517.zip added to bazel-mirror.


bazel-io:

hochhaus:
I think I originally missed that point of your comment. Were you saying that dl.google.com availability is good enough so no need for bazel-mirror?

If so, I think this change is ready to go.

hochhaus:
Also, I needed to disable two tests to work around an [already fixed](https://github.com/google/closure-library/commit/42fcead73bc36dc6586e5cbda1fe8b4416a785e6) library bug.

What is the best practice for temporarily disabling broken tests?

closure/compiler/BUILD
```diff
@@ -18,5 +18,5 @@ java_binary(
     name = "compiler",
     jvm_flags = ["-client"],
     main_class = "com.google.javascript.jscomp.CommandLineRunner",
-    runtime_deps = ["@closure_compiler//jar"],
+    runtime_deps = ["@closure_compiler//:compiler.jar"],
```
jart: Wow that actually works? You don't need to have a `java_import`?

closure/compiler/closure_compiler.BUILD
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+exports_files(["compiler.jar"])
```
jart: Since we're defining the build file anyway, mind defining a [java_import](http://bazel.io/docs/be/java.html#java_import) for this?

closure/library/BUILD
```diff
@@ -27,6 +27,7 @@ closure_js_library(
         "JSC_INVALID_SUPPRESS",
         "JSC_MISSING_JSDOC",
         "JSC_MISSING_RETURN_JSDOC",
+        "JSC_MISSING_REQUIRE_CALL_WARNING",
```
jart: [Optional] Consider filing an issue against Closure Library for this. Suppressions should be for legacy stuff. They shouldn't be introducing new ones.

closure/templates/BUILD
```diff
@@ -39,7 +39,6 @@ closure_js_library(
     name = "soyutils_usegoog",
     srcs = ["@soyutils_usegoog//file"],
     suppress = [
-        "JSC_MUST_HAVE_TRAILING_UNDERSCORE",
```
jart: Nice! I'm so happy that maintainability feature is coming in handy :D

closure/testing/test/BUILD
```diff
@@ -31,28 +31,31 @@ closure_js_library(
...
-    name = "arithmetic_scope_test",
-    srcs = ["arithmetic_scope_test.js"],
-    language = "ECMASCRIPT6_STRICT",
-    deps = [
-        ":arithmetic_module_lib",
-        "//closure/library",
-        "//closure/library:testing",
-    ],
-)
+# TODO(hochhaus): Enable test when upstream local file loader fix is included
```
jart: If this release introduces a bug that breaks our own tests, then is it going to be safe for our users? Shouldn't we wait another few days until they cut another release that fixes this awful regression?

jart:
Wow I totally didn't notice that Closure Compiler cut a new release last Wednesday. I guess I made the classic blunder of checking the branch list rather than the tag list :\

closure/library/BUILD
hochhaus: I just inspected these and something more is going on here. Each instance in the library has a suppress annotation on the function body.

[closure/goog/datasource/expr.js](https://github.com/google/closure-library/blob/master/closure/goog/datasource/expr.js#L267)
[closure/goog/graphics/vmlelement.js](https://github.com/google/closure-library/blob/master/closure/goog/graphics/vmlelement.js#L106)
[closure/goog/net/browsertestchannel.js](https://github.com/google/closure-library/blob/master/closure/goog/net/browsertestchannel.js#L349)
[closure/goog/net/channelrequest.js](https://github.com/google/closure-library/blob/master/closure/goog/net/channelrequest.js#L1122)

```
INFO: From Checking 812 JS files in //closure/library:library:
external/closure_library/closure/goog/datasource/expr.js:276: WARNING lintChecks JSC_MISSING_REQUIRE_CALL_WARNING - missing require: 'goog.ds.DataManager'
                                  goog.ds.DataManager.getInstance();
                                  ^

external/closure_library/closure/goog/graphics/vmlelement.js:116: WARNING lintChecks JSC_MISSING_REQUIRE_CALL_WARNING - missing require: 'goog.graphics.VmlGraphics'
      goog.graphics.VmlGraphics.toSizeCoord(height);
      ^

external/closure_library/closure/goog/graphics/vmlelement.js:276: WARNING lintChecks JSC_MISSING_REQUIRE_CALL_WARNING - missing require: 'goog.graphics.VmlGraphics'
  style.height = goog.graphics.VmlGraphics.toSizePx(height);
                 ^

external/closure_library/closure/goog/graphics/vmlelement.js:421: WARNING lintChecks JSC_MISSING_REQUIRE_CALL_WARNING - missing require: 'goog.graphics.VmlGraphics'
  style.height = goog.graphics.VmlGraphics.toPosPx(height);
                 ^

external/closure_library/closure/goog/net/browsertestchannel.js:376: WARNING lintChecks JSC_MISSING_REQUIRE_CALL_WARNING - missing require: 'goog.net.BrowserChannel'
      goog.net.BrowserChannel.createChannelRequest(this, this.channelDebug_);
      ^

external/closure_library/closure/goog/net/channelrequest.js:1129: WARNING lintChecks JSC_MISSING_REQUIRE_CALL_WARNING - missing require: 'goog.net.BrowserChannel'
  this.watchDogTimerId_ = goog.net.BrowserChannel.setTimeout(
                          ^

0 error(s), 6 warning(s)
```

Is it possible that JSChecker is not honoring suppress annotations on function bodies?

closure/templates/BUILD
hochhaus: This feature you added is amazing. I certainly would have missed removing it without the error.

jart: 💃

closure/library/BUILD
jart: It's very possible. Mind copying this into an issue and leaving a TODO?

closure/testing/test/BUILD
hochhaus: It isn't safe in the event they load files off of disk. For production deployments it shouldn't be an issue. It is a great point that this is a serious problem and it would be good to isolate our users.

Is a new release planned soon? If so, waiting is a great option. If not, we could also apply the loader bugfix from a patch file.

jart: Even if production is ok, we still want our unit testing to work. Can you tell me more about why this is broken? So we can gauge whether or not it's worth raising a fuss for them to cut another release?

jart:
> Were you saying that dl.google.com availability is good enough so no need for bazel-mirror in this case?

Correct.

closure/testing/test/BUILD
hochhaus: https://github.com/google/closure-library/commit/6453bc5c9d526120f759266b3f4d558511a54a92 added ES6 debug loader transpilation. However, in [goog.loadFileSync_](https://github.com/google/closure-library/commit/6453bc5c9d526120f759266b3f4d558511a54a92#diff-c359099ba4bc09bc2864536c49356d52R1477) it added a bug by assuming only "200" HTTP status codes from the XHR signifies success. This works for remote files, but fails for "file://" URLs (which return status of "0" on success). The [fix](https://github.com/google/closure-library/commit/42fcead73bc36dc6586e5cbda1fe8b4416a785e6#diff-c359099ba4bc09bc2864536c49356d52R1490) just relaxes the check to `xhr.status == 0 || xhr.status == 200`.

Therefore this is an issue when running in un-compiled mode or otherwise loading off of the local file system.

What is your preferred next step?

closure/library/BUILD
hochhaus: Absolutely. Filed https://github.com/bazelbuild/rules_closure/issues/73.

closure/testing/test/BUILD
hochhaus: I'm looking into this issue further and I think my explanation is wrong. More soon...

Please hold off on looking more so that you don't waste your time.

hochhaus: I think that setting goog.ENABLE_DEBUG_LOADER = false in phantomjs_runner.js along with a strict goog.module/goog.require behavior was the real cause of the test failure.

https://github.com/google/closure-library/wiki/goog.module:-an-ES6-module-like-alternative-to-goog.provide#how-do-i-use-a-googmodule-from-a-traditional-closure-file

I think I have addressed all issues with this PR now but I will check again in the morning when I have a fresh set of eyes.

Please hold off on review for now...

hochhaus:
I dug deeper into the goog.loadModule() issue and filed a new bug report https://github.com/google/closure-compiler/issues/1815. What I said previously about https://github.com/google/closure-library/commit/42fcead73bc36dc6586e5cbda1fe8b4416a785e6 being the source of the module loading test failures is incorrect. 

The file loader bug possibly isn't a show stopper. I frequently use "--allow-file-access-from-files" on my development browser so I ran into the issue almost immediately and incorrectly assumed that it was causing the test failures. I'm not sure how common loading files from "file://" URLs is for development. Using a local webserver for uncompiled files avoids the issue. How would you like to proceed?

PTAL

hochhaus:
Updated JsCheckerPassConfig.java as well.

jart:
`closure_js_test` starts a local webserver which hosts files under `/filez/`. So no `file://` URLs are needed. I would like to be able to run tests in raw sources mode, using the debug loader. But https://github.com/ariya/phantomjs/issues/14028 prevents that. That's the reason why everything has to be compiled in `WHITESPACE_ONLY` mode.

jart:
Jenkins: test this please

jart:
One more nit: I would personally prefer no periods at end of commit status lines in the future. See: http://chris.beams.io/posts/git-commit/

hochhaus:
> One more nit: I would personally prefer no periods at end of commit status lines in the future.

Absolutely. Good advise. Thanks.

Issue #67, opened 2016-05-20
Make closure_js_test compiled and more explicit


Dominator008:
Type safety should be encouraged in tests, https://github.com/bazelbuild/rules_closure#no-type-safety sounds misleading and is probably not the best practice.

Additionally, `closure_js_test` should be able to run both uncompiled and compiled. This can help finding Closure Compiler bugs.

jart:
I agree. But it's not practical, because the Closure Library testing utilities don't work with type checking. If you want to mock something, you have to use a zillion `/** @type {foo} */` comments in your functions. It's a nightmare.

I would love nothing more than to make tests always run in uncompiled mode. They would run so fast that way. But a bug in PhantomJS is blocking this at the moment: https://github.com/ariya/phantomjs/issues/14028 It seems to be the biggest bug PhantomJS is facing. Yet no one has fixed it for years. It's probably not an easy problem to solve. Maybe I'll take a crack at it some day. But I probably won't have time anytime soon.

Dominator008:
Then if tests don't use mocks we can have type safety?

Uncompiled by default is fine by me. I was talking about a parameter like `compiled = True` to allow tests to run both compiled and uncompiled.

jart:
We could add a parameter like that. The user would be required to use `goog.testing.testSuite` rather than defining `testFoo` functions. Would that be an acceptable tradeoff to you?

MatrixFrog:
"Mocks and typechecking don't play well together" is undeniably true but I definitely recommend that from that, you draw the conclusion "Avoid mocks when possible" and not "Don't bother typechecking your test code"

jart:
In that case @MatrixFrog, would you argue that type checking should be the default mode for `closure_js_test`?

MatrixFrog:
I think so. As you probably know, switching the default for the similar Google-internal rule will likely never happen because there's just too much non-typecheckable test code already there. But if there's going to be a new set of rules then it would probably be nice to set that as the default.

Dominator008:
@jart For running tests compiled, I assume `goog.require('goog.testing.jsunit');` also works? I don't mind explicitly adding a `goog.require` statement. In fact I prefer the explicitness so that it's clear that the assert functions are defined.

MatrixFrog:
I would agree, this seems like a clear "explicit is better than implicit" kind of situation. This is why our linter has a special exception for  goog.testing.jsunit

jart:
[...]

PR #75, opened 2016-05-27
Use full compilation for closure_js_test

jart:
I've addressed the feedback left by @Dominator008 and @MatrixFrog in #67. This was basically accomplished by changing `closure_js_test` to be a build macro rather than a Skylark rule. JsChecker updates come later. Please take a look.

The only difficulty I've really encountered is that `goog.testing.testSuite` keys need to be quoted. I would have assumed that since the type signature says they're strings, the `ADVANCED` property renamer wouldn't rename them? Also no one seems to quote them internally at Google. Is this intended behavior?

CC: @hochhaus


MatrixFrog:
Tests should set CompilerOptions.exportTestFunctions = true. There's a flag for that in the google-internal runner, but it looks like we don't have one externally.

README.md
```diff
@@ -444,15 +423,58 @@ path.
...
+
+### Arguments
+
+- **name:** ([Name][name]; required) Unique name for this rule.
+
+- **deps:** (List of [labels][labels]; required) Labels of Skylark rules
+  exporting `transitive_js_srcs`. Each source will be inserted into the webpage
+  in its own `<script>` tag based on a depth-first preordering.
+
+- **harness:** (Label; required; default is
```
hochhaus: optional

README.md
```diff
@@ -444,15 +423,58 @@ path.
...
-- **deps:** (List of [labels][labels]; optional) Direct dependency list. This
-  has the same meaning as it does in `closure_js_binary`.
+- **deps:** (List of [labels][labels]; optional) Direct dependency list passed
+  along to `closure_js_library`. This list will almost certainly need to include
+  `"@io_bazel_rules_closure//closure/library:testing"`.
 
-- **language:** (String; optional; default is `"ECMASCRIPT5_STRICT"`) Variant of
-  JavaScript in which `srcs` are written. See the `closure_js_library`
-  documentation for more information.
+- **html:** (Label; required; default is
```
hochhaus: optional

closure/private/defs.bzl
```diff
@@ -42,16 +43,20 @@ JS_HIDE_WARNING_ARGS = [
     "--hide_warnings_for=.soy.js",
     "--hide_warnings_for=external/closure_library/",
     "--hide_warnings_for=external/soyutils_usegoog/",
+    "--hide_warnings_for=external/protobuf_js/",
```
hochhaus: Nice. This is helpful.

closure/testing/closure_js_test.bzl
```diff
@@ -14,128 +14,56 @@
...
+                    html=None,
+                    compilation_level="ADVANCED",
+                    css=None,
+                    defs=[],
+                    dependency_mode="LOOSE",
+                    entry_points=[],
+                    formatting="PRETTY_PRINT",
+                    language=JS_LANGUAGE_DEFAULT,
+                    pedantic=False,
+                    suppress=[]):
```
hochhaus: [Nit] Inconsistent spacing around equals signs. (I know I do this too, maybe I should cleanup the entire rules_closure repo in a single commit).

hochhaus: For the defaults, I found the following section of the bazel [manual](http://bazel.io/docs/skylark/macros.html) interesting:

```
In most cases, optional parameters should have a default value of None. None
can be passed directly to native rules, which treat it the same as if you had not
passing any argument. Thus, there is no need to replace it with 0, False, or [] for
this purpose. Instead, the macro should defer to the rules it creates, as their
defaults may be complex or may change over time. Additionally, a parameter
that is explicitly set to its default value looks different than one that is never set
(or set to None) when accessed through the query language or build-system
internals.
```

What are your thoughts here?

closure/private/defs.bzl
jart: Note: When I write a compiler frontend for `closure_js_binary` (similar to JsChecker) we're going to have a better way of dealing with this—via the suppression system. I'm starting to realize now that we're going to have to propagate suppressions to parent rules on a per-file basis.

hochhaus: > I'm starting to realize now that we're going to have to propagate suppressions
> to parent rules in a way that associates them with groups of files.

This sounds amazing!

README.md
```diff
@@ -1,9 +1,9 @@
 # Closure Rules for Bazel (αlpha) [![Build Status](https://travis-ci.org/bazelbuild/rules_closure.svg?branch=master)](https://travis-ci.org/bazelbuild/rules_closure)
 
-JavaScript | Templating | Stylesheets | Protocol Buffers
---- | --- | --- | ---
+JavaScript | Templating | Stylesheets | Miscellaneous
+--- | --- | --- | --- | ---
```
hochhaus: Is this too many columns?

hochhaus:
I reviewed all the code and don't see anything major for improvement. Fully compiled tests are really cool!

LGTM

closure/private/defs.bzl
```diff
@@ -42,16 +43,20 @@ JS_HIDE_WARNING_ARGS = [
     "--hide_warnings_for=.soy.js",
```
hochhaus: Can you remove ".soy.js" here? I think all .soy.js files live under bazel-out/local-fastbuild/genfiles/.

closure/testing/closure_js_test.bzl
jart: That's amazing. I'm so happy you knew about this!

jart:
All comments addressed.

@MatrixFrog Thanks for letting me know about `exportTestFunctions`. I'm going to look into creating a Java frontend for JSComp soon, so I can hopefully enable things like that. I can also send a CL back upstream to you for enabling that flag in the open source world at some point in the near future.

PR #78, opened 2016-05-27
Add common defines and output_wrapper to closure_js_binary()

hochhaus:
I find myself adding certain boilerplate in nearly every `closure_js_binary()`. What do you think of setting up a few common defines (`goog.STRICT_MODE_COMPATIBLE`, `goog.soy.REQUIRE_STRICT_AUTOESCAPE`) and an output wrapper automatically?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Closure Rules for Bazel (αlpha) [![Build Status](https://travis-ci.org/bazelbuild/rules_closure.svg?branch=master)](https://travis-ci.org/bazelbuild/rules_closure)
 
 JavaScript | Templating | Stylesheets | Miscellaneous
---- | --- | --- | --- | ---
+--- | --- | --- | ---
 [closure_js_library](#closure_js_library) | [closure_template_js_library](#closure_template_js_library) | [closure_css_library](#closure_css_library) | [closure_proto_js_library](#closure_proto_js_library)
 [closure_js_binary](#closure_js_binary) | [closure_template_java_library](#closure_template_java_library) | [closure_css_binary](#closure_css_binary) | [phantomjs_test](#phantomjs_test)
 [closure_js_deps](#closure_js_deps) | [closure_template_py_library](#closure_template_py_library) | |
@@ -278,7 +278,8 @@ target. See the documentation of the `deps` attribute for further information.
 ```python
 load("@io_bazel_rules_closure//closure:defs.bzl", "closure_js_binary")
 closure_js_binary(name, deps, css, pedantic, debug, language, entry_points,
-                  dependency_mode, compilation_level, formatting, defs)
+                  dependency_mode, compilation_level, formatting,
+                  output_wrapper, defs)
 ```
 
 Turns JavaScript libraries into a minified optimized blob of code.
@@ -375,6 +376,13 @@ This rule must be used in conjunction with `closure_js_library`.
   - `PRETTY_PRINT`
   - `PRINT_INPUT_DELIMITER`
   - `SINGLE_QUOTES`
+
+- **output_wrapper:** (String; optional; default is
+  `"(function(){%output%}).call(this);"`) Wraps the compiled code. The
+  [default][output-wrapper-faq] specifies an anonymous function to prevent
+  variables from polluting the global scope. If you are using compiler level
+  modules you should disable this option with `output_wrapper=""` and pass
+  `module_wrapper` via `defs` instead.
 
 - **defs:** (List of strings; optional) Specifies additional flags to be passed
   to the Closure Compiler, e.g. `"--hide_warnings_for=some/path/"`. To see what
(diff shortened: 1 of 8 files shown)


bazel-io:

README.md
```diff
@@ -375,6 +376,14 @@ This rule must be used in conjunction with `closure_js_library`.
   - `PRINT_INPUT_DELIMITER`
   - `SINGLE_QUOTES`
 
+- **output_wrapper:** (String; optional; default is
+  `"(function(){%output%}).call(this);"`) Wraps the compiled code. The
+  [default](https://github.com/google/closure-compiler/wiki/FAQ#when-using-advanced-optimizations-closure-compiler-adds-new-variables-to-the-global-scope-how-do-i-make-sure-my-variables-dont-collide-with-other-scripts-on-the-page)
```
jart: Please put long link at the bottom of the document.

closure/compiler/closure_js_binary.bzl
```diff
@@ -67,12 +68,20 @@ def _impl(ctx):
     args += ["--debug"]
   elif is_using_closure_library(srcs):
     args += ["--define=goog.DEBUG=false"]
+  if is_using_closure_library(srcs) and language_out in ["ECMASCRIPT6_TYPED",
```
jart: Mind making a `_STRICT_LANGUAGES` set?

closure/compiler/closure_js_binary.bzl
```diff
@@ -67,12 +68,20 @@ def _impl(ctx):
...
+                                                         "ECMASCRIPT6_STRICT",
+                                                         "ECMASCRIPT5_STRICT"]:
+    args += ["--define=goog.STRICT_MODE_COMPATIBLE"]
+  if contains_file(srcs, ctx.attr._soyutils_usegoog):
```
jart: This is meant to take a hard-coded string path.

closure/compiler/closure_js_binary.bzl
```diff
@@ -67,12 +68,20 @@ def _impl(ctx):
...
+    args += ["--define=goog.soy.REQUIRE_STRICT_AUTOESCAPE"]
   for entry_point in ctx.attr.entry_points:
     _validate_entry_point(entry_point, srcs)
     args += ["--entry_point=" + entry_point]
   if ctx.attr.pedantic:
     args += JS_PEDANTIC_ARGS
     args += ["--use_types_for_optimization"]
+  if ctx.attr.output_wrapper:
+    args += ["--output_wrapper=\"%s\"" % ctx.attr.output_wrapper]
```
jart: Is it safe to pass this without the quotes? Can we have a file_test in `closure/compiler/test/` somewhere that tests an output_wrapper witha a space?

closure/compiler/test/BUILD
```diff
@@ -43,7 +43,7 @@ file_test(
                "\"version\":3,\n" +
                "\"file\":\"/closure/compiler/test/hello_bin.js\",\n" +
                "\"lineCount\":1,\n" +
-               "\"mappings\":\"AAgBAA,OAAAC,IAAA,CAAgB,aAAhB;\",\n" +
```
jart: That's really good that the compiler is smart enough to take the wrapper into consideration when generating the sourcemap.

closure/compiler/test/BUILD
```diff
@@ -33,7 +33,7 @@ closure_js_binary(
 
 file_test(
     name = "minification",
-    content = "console.log(\"hello world\");\n",
+    content = "(function(){console.log(\"hello world\");}).call(this);\n",
```
jart: Please update one of these tests with `output_wrapper = "%output%"`.

jart:
These are all great catches. Thank you so much.

Once I LGTM, would it be too much trouble to ask you to break this down into three separate commits within the PR?

- Make strict escaping mandatory on Closure Templates
- Define STRICT_MODE_COMPATIBLE if combination is strict
- Add output_wrapper flag to closure_js_binary

Then I'll do a traditional merge.

That should make my life a little easier when I cut the release.

jart:
Also rebase please.

closure/compiler/closure_js_binary.bzl
hochhaus: Good point.

As the filename changes with the version (eg: `soy-2016-01-12-soyutils_usegoog.js`) I was attempting to abstract it via the label. I'm fixing the bug in the next patch set (still via the label though). Please let me know if you prefer a different solution. I'm happy to make another revision.

closure/compiler/test/BUILD
hochhaus: Are you looking to disable the output_wrapper for this test? If so, do you prefer `output_wrapper = None`, `output_wrapper = ""` or `output_wrapper = "%output%"`?

jart: I'm mostly just looking for a test that uses the new flag. It might be cool to demonstrate to the user how to disable the output wrapper entirely. Of the three you listed, which one is canonical? And if multiple ones do the same thing, can we have a fail() check that forces the user to choose only one of them?

closure/compiler/closure_js_binary.bzl
jart: Ok that makes sense. I guess it wouldn't be so terrible to have such a tiny dependency. But in that case, you probably want to change `_soyutils_usegoog` to have `single_file=True` and then this parameter should be `ctx.file._soyutils_usegoog.short_path` to avoid the type error.

I would recommend writing a failure test to verify your change is correct, but I haven't gotten around to adding an `internal_expect_failure` flag to `closure_template_js_library` yet, which is currently only available on `closure_js_library`.

jart: Another alternative if you prefer, would possibly be to just write a one-off method that says: `for src in srcs: if src.startswith("external/soyutils_usegoog/") and src.endswith("soyutils_usegoog.js"): ...`

hochhaus:
Comments address and rebased.

PTAL

jart:
Mr. Jenkins: test this please

jart:
LGTM

PR #85, opened 2016-06-01
Enforce style with clang-format

hochhaus:
Unfortunately, I think the design of this change will still require a fair bit of discussion. I'm not sure the best way to lay in the style enforcement.

The source of the issue is that clang-format's intended use case is to automatically re-format files into the correct syntax. This is amazing when clang-format is run from your editor's save hook. However, I wanted something a bit stricter than just relying on each team member to configure their editor correctly. Therefore, this change breaks the compile if any `*.{js,proto}` file requires reformatting. However, it will not automatically reformat the source files. (This is good I think since changing the _inputs_ to the build during the build seems horrible.) Instead the user must invoke clang-format (either from their editor or from the command line) to reformat the file.

I think at least the following items need discussion during the merge:
- I added emacs specific documentation since that is the only editor I know well enough to configure. Do we need docs for other editors prior to merge? Is editor specific documentation acceptable?
- In the case where clang-format is run automatically at save from the editor `style = True` should never catch any errors. We hope this to be the common case. What should the error say/do when this isn't the case? What is the correct way to return a nice error message to the user here? Just "exit 1"?
- Is referencing external MELPA sources for editor configuration acceptable?
- What is the correct way to setup the editor to locate the clang-format binary? Referencing the copy in external/ seem sloppy since the user could have multiple clones of different repos which all depend on rules_closure. Alternatively, having the user install a separate copy also seems messy since thje version might skew with the rules_closure managed version.

### DIFF
--- a/closure/compiler/closure_js_library.bzl
+++ b/closure/compiler/closure_js_library.bzl
@@ -65,7 +65,12 @@ def _impl(ctx):
     for edep in direct_dep.js_exports:
       args += ["--dep=%s" % edep.js_provided.path]
       inputs.append(edep.js_provided)
-  args += ["--suppress=%s" % s for s in ctx.attr.suppress]
+  clangFormat = True
+  for suppress in ctx.attr.suppress:
+    if suppress == "CLANG_FORMAT":
+      clangFormat = False
+    else:
+      args += ["--suppress=%s" % suppress]
   if ctx.attr.internal_expect_failure:
     args += ["--expect_failure"]
   if ctx.attr.internal_expect_warnings:
@@ -74,6 +79,25 @@ def _impl(ctx):
                          "%s_worker_input" % ctx.label.name)
   ctx.file_action(output=argfile, content="\n".join(args))
   inputs.append(argfile)
+
+  files = [ctx.outputs.provided, ctx.outputs.stderr]
+  if clangFormat and len(ctx.files.srcs) > 0:
+    files.append(ctx.outputs.jsstyle)
+    style_args = [ctx.executable._clangformat.path,
+                  ctx.outputs.jsstyle.path]
+    style_inputs = [ctx.executable._clangformat]
+    for direct_src in ctx.files.srcs:
+      style_args += [direct_src.path]
+      style_inputs.append(direct_src)
+    ctx.action(
+        inputs=style_inputs,
+        outputs=[ctx.outputs.jsstyle],
+        executable=ctx.executable._googleformat,
+        arguments=style_args,
+        mnemonic="JSStyleChecker",
+        progress_message="Checking %d JS files for style in %s" % (
+            len(ctx.files.srcs) + len(ctx.files.externs), ctx.label))
+
   ctx.action(
       inputs=inputs,
       outputs=[ctx.outputs.provided, ctx.outputs.stderr],
@@ -101,6 +125,15 @@ def _determine_check_language(language):
     return "ECMASCRIPT3"
   return language
 
+def _outputs(attr):
+  outputs = {
+      "provided": "%{name}-provided.txt",
+      "stderr": "%{name}-stderr.txt",
+  }
+  if "CLANG_FORMAT" not in attr.suppress and len(attr.srcs) > 0:
+    outputs["jsstyle"] = "%{name}-jsstyle.txt"
+  return outputs
+
 closure_js_library = rule(
     implementation=_impl,
     attrs={
@@ -122,8 +155,11 @@ closure_js_library = rule(
             executable=True),
         "_closure_library_base": CLOSURE_LIBRARY_BASE_ATTR,
         "_closure_library_deps": CLOSURE_LIBRARY_DEPS_ATTR,
+        "_googleformat": attr.label(
+            default=Label("//third_party/llvm/llvm/tools/clang:google_format"),
+            executable=True),
+        "_clangformat": attr.label(
+            default=Label("//third_party/llvm/llvm/tools/clang:clang_format_extract"),
+            executable=True),
     },
-    outputs={
-        "provided": "%{name}-provided.txt",
-        "stderr": "%{name}-stderr.txt",
-    })
+    outputs=_outputs)
--- a/README.md
+++ b/README.md
@@ -112,6 +112,39 @@ maven_jar(
 )
 ```
 
+### Source code formatting
+
+It is recomended that you configure your editor to format your JavaScript and
+Protocol Buffer source files automatically at save. Doing so makes it easy to
+follow Google coding style conventions.
+
+#### Emacs
+
+This section configures `js2-mode` as your major mode for JavaScript editing
+and uses `clang-format` to format JavaScript and Protocol Buffer source files
+automatically when saving. Depending on your preferences you may need to
+customize these instructions. You will also need to ensure that `clang-format`
+is in your `PATH`.
+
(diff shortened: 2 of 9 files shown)


bazel-io:

jart:
To make the review process simpler, could you limit this pull request to to simply adding support for clang format? Then we can do a follow-up review for integrating it into the existing rules.

jart:
Also regarding emacs, have you seen: https://github.com/jart/js2-closure ?

hochhaus:
I recently found it! It looks so awesome I created Issue #86 this morning to consider integrating it.

The awesomeness of clang-format/js2-closure reminds me of how smooth gofmt/goimports is for writing go.

closure/compiler/test/hello.js
```diff
@@ -14,4 +14,4 @@
 
 // hello world
 
-console . log ( 'hello world' ) ;
```
jart: Please revert this change and add a `// clang-format off` comment.

closure/compiler/closure_js_library.bzl
```diff
@@ -68,6 +68,25 @@ def _impl(ctx):
...
+  files = [ctx.outputs.provided, ctx.outputs.stderr]
+  if ctx.attr.style and len(ctx.files.srcs) > 0:
+    files.append(ctx.outputs.jsstyle)
+    style_args = [ctx.executable._clangformat.path,
+                  ctx.outputs.jsstyle.path]
+    style_inputs = [ctx.executable._clangformat]
+    for direct_src in ctx.files.srcs:
+      style_args += [direct_src.path]
+      style_inputs.append(direct_src)
+    ctx.action(
```
jart: When we do the `closure_js_library` integration in a follow-up pull request, do you think it would be better if we launched clang format as a concurrent subprocess of JsChecker? The advantage is that with a significant amount of kludgery, we might be able to integrate it into JsChecker's error system. Therefore clang format errors could be turned off via suppress rather than a new attribute. How feasible does that sound, based on your understanding of clang format so far?

hochhaus: >  do you think it would be better if we launched clang format as a concurrent subprocess of JsChecker?

I'm fine doing it that way if you prefer.

One downside is that such a job would be hidden to "bazel query" and therefore is slightly more magical to new users of rules_closure who are coming from a bazel background.

> we might be able to integrate it into JsChecker's error system.

Are you thinking a file wide fileoverview level @suppress or something more granular (eg: function, line level)?

One downside to non-standard clang-format suppressions is that existing clang-format users (eg: C/C++/Java/Objective-C) won't expect it. 

> Therefore clang format errors could be turned off via suppress rather than a new attribute.

Oh, maybe you are thinking a `closure_js_library` level suppression (vs an in-file suppression)? Is that correct?

If so, couldn't we implement that API in `closure_js_library` either via a conditional ctx.action or a JsChecker subprocess? (Basically eat the clangFormat suppression in `closure_js_library` , do not run the ctx.action for clang-format and do not pass the suppression down to JsChecker).

> How feasible does that sound, based on your understanding of clang format so far?

If you are thinking about avoiding a new `style` attribute and conditionally running clang-format when the suppression is not present (either from a conditional ctx.action or a JsChecker subprocess) this sounds very actionable.

If you are thinking of fine grained in-file suppressions it would potentially be complex. I would need to look at the clang-format source to gain an idea of the complexity.

jart: I'm only talking about using the `convention` and `suppress` attributes on `closure_js_library` to configure the clang style enforcement. I think the convention using the `// clang-format off` comments should stay the same, rather than using `@suppress`.

Also since we're already having design discussion here about enforcement via closure_js_library, I would recommend creating a separate PR for the simple task of adding the formatter to the codebase. That's something we can get merged in quicker.

hochhaus: > I'm only talking about using the convention and suppress attributes on closure_js_library to configure the clang style enforcement.

Thanks, this plan sounds great to me.

> I would recommend creating a separate PR for the simple task of adding the formatter to the codebase.

Coming very soon.

hochhaus:
Please hold on review of this PR until https://github.com/bazelbuild/rules_closure/pull/92 merges. I moved to the suppress based API in this updated code (expect for the protobuf case which does not have any suppressions). We could add suppress to the protobuf rule, but that is also a bit funky. The lack of API symmetry between JS and PB makes me wonder if this is the best course of action.

I have no strong feelings on the manner to disable style checks. If you let me know what you think is best I will implement it.

jart:
Now that we've got the previous change merged, let's think more about what we can do to get this one in the codebase.

The way style / linting enforcement has traditionally been done inside Google is we have a `METADATA` file for each project. It contains a protobuf message that configures the presubmit system. This runs the linting / style enforcement code before changelists get mailed out. Google has never integrated style checks directly into the standard Blaze rules themselves, as far as I know.

But I feel like there's a lot of value to the unified, integrated direction we've been moving in. I think adding the linter was a great decision. There was very little cost to doing so. But adding the heavyweight clang check to the JS library rule means things are getting serious. There's >100k js_library rules in Google's codebase. So integrating might make these rules harder to universalize internally. Also with this design, if `style = False` the Clang binary is still linked in. The user still has to pay the price of the download for something they're not using. Therefore it might be better to perform this check as a separate rule.

Do you think it might be better if we try going the separate rule route first? And then later decide if it should be merged in, after the socializing the idea with more people?

hochhaus:
I finally have time to look at this change again. :)

In terms of reducing dependency weight, I have two ideas:

* Since we are bazel-mirroring the clang package anyways, pull out just the clang-format binary and host that in a dedicated location. This reduces the download size from ~146M to 514K (xz compressed). This is helpful for users who want style checking.
* Add a macro wrapper around `closure_js_libarary()` which delegates to the `_closure_js_library()` rule for the majority of the work. However, from the macro, conditionally `native.genrule()` a job which enforces the style formatting _only_ when `style=1`. This let's us avoid the dependency on clang-format all-together when `style=0`.

I want to make sure that this change does not make it harder to deploy `closure_rules` universally internally.

In combination with the above proposals to minimize the weight of the dependencies, would disabling style checking by default address your concerns?

jart:
What if we created a new bazelbuild/rules_style project that used clang-format, buildifier, google-java-format, etc. all rolled up into one rule? @damienmg 

# Style Rules Design Doc

Build macro for enforcing style.

## Usage

User puts the following in each build file:

```python
style_test(
    name = "style",
    files = glob(["**"]),
)
```

They get invoked when user runs routine `bazel test //...`. Rule runs formatter on each file and fails if output of formatter isn't identical to file. Displays diff. Prints command that can be copy/pasted to auto-format the file.

## Implementation

External repositories file has all the major tools, e.g. clang-format, google-java-format, buildifier, etc. The name of the game is to only get the user to pay the price of downloading what he needs to validate.

Therefore each language is implemented as a separate Skylark rule. These rules are private. They are tied together by a master macro that considers file extensions.

The master macro might look as follows:

```python
def style_test(name, files, **kwargs):
  java_files = [f for f in files if f.endswith('.java')]
  if java_files:
    _style_java_test(name = name + '_java', files = java_files, **kwargs)
  # ...
```

hochhaus:
@jart I'm a fan of the rules_style project.

jart:
I'm going to close this out for the time being, until we can figure out how we want to approach this.

PR #90, opened 2016-06-01
Rename protobuf rules to closure_<lang>_proto_library format

hochhaus:


bazel-io:

jart:
Can you add a third commit providing a backwards compatible rule for the template ones only? E.g.

```python
def closure_template_java_library(**kwargs):
  print("Deprecated: use closure_java_template_library() instead, " +
        "closure_template_java_library will be removed in version 0.3.0")
  closure_java_template_library(**kwargs)
```

jart:
Also would it be possible to break this into two pull requests?

jart:
Mr. Jenkins: test this please

jart:
LGTM. Thanks so much for sending this.

Issue #77, opened 2016-05-27
How is this different/same as plovr?


pcj:
I'm just looking at this project and trying to figure out if it replaces plovr.  At first glance it seems like there is overlap with this project and what @bolinfest had intended many years ago with plovr.  

One of things that I can't give up with plovr is build times.  the `plovr serve` command is the only good way I know of keeping a JVM hot with the closure compiler.  With a cold start JVM my build times are on the order of 30s, but with plovr serve after the first compile they're about 4-5s.  

Does this project somehow replicate the `plovr serve` command? Again the crux issue is keeping the JVM running between compiles.

pcj:
Also, are groups within google using this?

hochhaus:
Good question about minimizing compile times via a hot JVM.

> Also, are groups within google using this?

With the [0.1.0 release](https://github.com/bazelbuild/rules_closure/releases/tag/0.1.0) @jart stated:

> This release brings Closure Rules to a level of quality that it's now been adopted by a 100k+ LOC codebase.

That is almost certainly a group within Google.

damienmg:
@pcj: Bazel is a curated version of the build system that is used inside Google. Blaze (bazel inside Google) has a >20k engineers daily active users.

Closure rules are a project started to replace the internal JS rules used inside Google (not open-sourced for a variety of reasons, mostly because they are too specific to how we do JS development at Google). @jart is a talented engineer from Google that developed those rules to work for the need of her team (and other teams). The closure rules provided for Bazel have reached such a level that they can safely be used to support a variety of project. As a lead of the Bazel project, I strongly encourage people to tries those rules out and tell if anything is wrong so those rules can be used outside of Google.

For keeping the JVM hot, we do have a worker strategy, See http://www.bazel.io/blog/2015/12/10/java-workers.html. I think this is not yet used for closure rules but it can definitely be used. @jart should talk to @philwo to use it.

jart:
@damienmg I'm really touched. Thank you for your kind words of support.

[...]

Issue #87, opened 2016-06-01
Bump to 0.1.1


pcj:
...or whatever your next release label is.  The README.md is a little behind: I was struggling to figure out why closure_proto_js_library could not be found until I realized it does not exist in 0.1.0.  Worked after I upgraded to: 

```
git_repository(
    name = "io_bazel_rules_closure",
    remote = "https://github.com/bazelbuild/rules_closure.git",
    commit = "80d493d5ffc3099372929a8cd4a301da72e1b43f"
)
```

jart:
Done. https://github.com/bazelbuild/rules_closure/releases/tag/0.2.0

https://groups.google.com/forum/#!topic/closure-rules-announce/7-iXVqx1EL8 <-- please subscribe to that list if you haven't already.

PR #95, opened 2016-06-02
Make JsChecker a Bazel worker

jart:
I'm following through on @damienmg's suggestion in #77.

This change significantly improves the performance of closure_js_library when `--strategy=JsChecker=worker` is passed to Bazel. Now it won't have to spin up a new JVM for every single library rule. This will be indispensable for projects that have one library rule for every js file. It also helps alleviates @pcj's concerns. However this only applies to library rules at the moment. I haven't implemented this for binary rules yet.

I copied some code from the Bazel codebase because I'm afraid to link against `@closure_tools//...` in the interest of hermeticism. I hope I'm doing things correctly.

CC: @philwo @hochhaus

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,4 @@ before_install:
   - ./bazel-0.2.2b-installer-linux-x86_64.sh --user
 
 script:
-  - bazel --batch --host_jvm_args=-Xmx1000m --host_jvm_args=-Xms1000m test closure/... --verbose_failures --test_output=errors --test_strategy=standalone --spawn_strategy=standalone --genrule_strategy=standalone --local_resources=1000,1,1.0
+  - bazel --batch --host_jvm_args=-Xmx1000m --host_jvm_args=-Xms1000m test closure/... --strategy=Javac=worker --strategy=JsChecker=worker --worker_verbose --verbose_failures --test_output=errors --test_strategy=standalone --spawn_strategy=standalone --genrule_strategy=standalone --local_resources=1000,1,1.0
--- a/closure/private/gensrcjar.sh
+++ b/closure/private/gensrcjar.sh
@@ -1,0 +1,85 @@
+#!/bin/sh
+#
+# Copyright 2016 The Closure Rules Authors. All rights reserved.
+# Copyright 2016 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# List of environment variables that must be defined as input.
+#
+# We use environment variables instead of positional arguments or flags for
+# clarity in the caller .bzl file and for simplicity of processing.  There is no
+# need to implement a full-blown argument parser for this simple script.
+INPUT_VARS="JAR OUTPUT PREPROCESSOR PROTO_COMPILER SOURCE"
+
+# Now set defaults for optional input variables.
+: "${PREPROCESSOR:=cat}"
+
+# Basename of the script for error reporting purposes.
+PROGRAM_NAME="${0##*/}"
+
+# A timestamp to mark all generated files with to get deterministic JAR outputs.
+TIMESTAMP=198001010000
+
+# Prints an error and exits.
+#
+# Args:
+#   ...: list(str).  Parts of the message to print; all of them are joined
+#       with a single space in between.
+err() {
+  echo "${PROGRAM_NAME}: ${*}" 1>&2
+  exit 1
+}
+
+# Entry point.
+main() {
+  [ ${#} -eq 0 ] || err "No arguments allowed; set the following environment" \
+      "variables for configuration instead: ${INPUT_VARS}"
+  for var in ${INPUT_VARS}; do
+    local value
+    eval "value=\"\$${var}\""
+    [ -n "${value}" ] || err "Input environment variable ${var} is not set"
+  done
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -21,4 +21,4 @@ before_install:
   - ./bazel-0.2.2b-installer-linux-x86_64.sh --user
 
 script:
-  - bazel --batch --host_jvm_args=-Xmx1000m --host_jvm_args=-Xms1000m test closure/... --verbose_failures --test_output=errors --test_strategy=standalone --spawn_strategy=standalone --genrule_strategy=standalone --local_resources=1000,1,1.0
+  - bazel --batch --host_jvm_args=-Xmx1000m --host_jvm_args=-Xms1000m test closure/... --strategy=Javac=worker --strategy=JsChecker=worker --worker_verbose --verbose_failures --test_output=errors --test_strategy=standalone --spawn_strategy=standalone --genrule_strategy=standalone --local_resources=1000,1,1.0
```
hochhaus: Do you want to suggest in the README that the user setup these strategies in a .bazelrc file?

jart: I figure we'll wait until the next release to document that. Or until we create a separate documentation site. Someone filed an issue yesterday about how the README docs were out of sync #87

closure/private/gensrcjar.sh
```diff
@@ -0,0 +1,85 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# List of environment variables that must be defined as input.
+#
+# We use environment variables instead of positional arguments or flags for
+# clarity in the caller .bzl file and for simplicity of processing.  There is no
+# need to implement a full-blown argument parser for this simple script.
```
hochhaus: Do you want to `set -e` and `set -u`?

```
-e  Exit immediately if a command exits with a non-zero status.
-u  Treat unset variables as an error when substituting.
```

.travis.yml
pcj: Nice!  Excited to try this out.

closure/private/java_proto_library.bzl
```diff
@@ -0,0 +1,76 @@
...
+      mnemonic="GenProtoSrcJar",
+      use_default_shell_env=True)
+  return struct(runfiles=ctx.runfiles(collect_default=True))
+
+_gensrcjar = rule(
+    implementation=_impl,
+    attrs = {
+        "src": attr.label(
+            allow_files = FileType([".proto"]),
+            single_file = True),
```
hochhaus: Can't multiple *.proto files be passed to the same rule?

.travis.yml
hochhaus: Good point.

We could also have a "dev" branch which we commit all development work to and then send it over to "master" only when we do the release. I'm fine with either solution.

java/com/google/javascript/jscomp/JsChecker.java
```diff
@@ -314,7 +319,11 @@ public CheckLevel level(JSError error) {
     return result.build();
   }
 
-  public static void main(String[] args) throws IOException {
+  static int processRequest(Collection<String> args) throws IOException {
+    String lastArg = Iterables.getLast(args, "");
+    if (lastArg.startsWith("@")) {
+      args = Files.readAllLines(Paths.get(lastArg.substring(1)), UTF_8);
+    }
```
hochhaus: Do you want to error if `lastArg.startsWith("@")` and more than one argument was passed?

java/com/google/javascript/jscomp/JsChecker.java
```diff
@@ -325,14 +334,58 @@ public static void main(String[] args) throws IOException {
...
+      }
+      WorkResponse.newBuilder()
+          .setOutput(buffer.toString())
+          .setExitCode(exitCode)
+          .build()
+          .writeDelimitedTo(System.out);
+      System.out.flush();
+      System.gc();  // be a good little worker process and consume less memory when idle
+    }
+  }
```
hochhaus: Very cool!

.travis.yml
jart: I spoke to one of the lords of documentation at Google and it seems like the canonical approach these days is to have a doc/ directory. You edit that with each change. Then when release time comes around, you run a tool to sync that to your `gh-pages` branch which [GitHub Pages](https://pages.github.com/) then renders to the website.

hochhaus: That sounds awesome.

hochhaus:
LGTM

closure/private/java_proto_library.bzl
jart: Not at the moment. This is a private scrap rule mostly copied in from the Bazel codebase. It's good enough for our needs. Provided there aren't any glaring bugs.

hochhaus: Fair enough.

closure/private/gensrcjar.sh
jart: This is pretty much copied verbatim from the Bazel codebase.

hochhaus: Sounds good to me. I understand the benefit in not forking the implementation.

java/com/google/javascript/jscomp/JsChecker.java
jart: In theory if that happens, it means the prior arguments were meant for the worker and can be safely ignored. I'm assuming you saw [ExampleWorker](https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/worker/ExampleWorker.java)? I changed it to be more forgiving here because I saw some error message in Bazel that specifically talked about the `@` thing relating to the last parameter.

hochhaus: Thanks, I actually had not seen the ExampleWorker yet. I agree your code is fine as-is.

philwo:
Looks great! :)

jart:
Thanks for the reviews guys!

In a future PR, I think it might actually be possible for me to write a generic wrapper around the main function of any Java program. I'll look into doing that.

PR #105, opened 2016-06-30
Add a "supports-workers" execution info tag to JsChecker.

philwo:
This is a new, upcoming requirement that helps Bazel decide whether an action can be executed via workers or not.


bazel-io:

jart:
Why closed?

philwo:
@jart Sorry, I used the wrong e-mail address for the commit, so the CLA check failed. I've resent it with the correct one. :)

PR #106, opened 2016-06-30
Add a "supports-workers" execution info tag to JsChecker.

philwo:
This is a new, upcoming requirement that helps Bazel decide whether an action can be executed via workers or not.


bazel-io:

jart:
Why does Bazel want `"1"` to signal `True`?

jart:
Mr. Jenkins: Test this please.

philwo:
@jart AFAIK the execution_info dict is a string -> string map, which means I can't use a boolean as the value. I picked "1", but it could have been "true", "yes" too.

PR #109, opened 2016-07-06
Automatically set --assume_function_wrapper

### LINKED ISSUE #101: Set --assume_function_wrapper when using a function output_wrapper
It would be nice if `--assume_function_wrapper` were automatically passed to the compiler when a proper output_wrapper was specified.

```
--assume_function_wrapper: Enable additional optimizations based
                                          on the assumption that the output
                                          will be wrapped with a function
                                          wrapper.  This flag is used to
                                          indicate that "global" declarations
                                          will not actually be global but
                                          instead isolated to the compilation
                                          unit. This enables additional
                                          optimizations.
```

As part of this issue, I wonder if we should discuss the default value for output_wrapper again. To me it seems like most code wants a default output_wrapper and `--assume_function_wrapper` supplied automatically. This does make `rules_closure` harder to use for some existing code (requiring an explicitly override to disable the output_wrapper). However, it seems like now (before the API is marked stable) is the correct time to make these backwards incompatible changes. I don't have strong feelings here but I thought I would mention it as a related issue.

hochhaus:
Fixed #101

### DIFF
--- a/closure/compiler/closure_js_binary.bzl
+++ b/closure/compiler/closure_js_binary.bzl
@@ -102,6 +102,8 @@ def _impl(ctx):
              "--define=goog.json.USE_NATIVE_JSON"]
   if ctx.attr.output_wrapper:
     args += ["--output_wrapper=%s" % ctx.attr.output_wrapper]
+    if ctx.attr.output_wrapper == "(function(){%output%}).call(this);":
+      args += ["--assume_function_wrapper"]
   if ctx.outputs.property_renaming_report:
     report = ctx.outputs.property_renaming_report
     files += [report]


bazel-io:

closure/compiler/closure_js_binary.bzl
```diff
@@ -101,6 +101,8 @@ def _impl(ctx):
              "--define=goog.json.USE_NATIVE_JSON"]
   if ctx.attr.output_wrapper:
     args += ["--output_wrapper=%s" % ctx.attr.output_wrapper]
+    if ctx.attr.output_wrapper == "(function(){%output%}).call(this);":
```
hochhaus: It isn't clear to me how many variants of output_wrappers we should try to support (eg: extra whitespace, etc). Also, if compiler level module support is later added we would need to support more complex module_wrapper patterns (eg: `var __sg__=__sg__||{};(function(a){%s}).call(this, __sg__);`). Currently the only compelling use case that I know of for output_wrappers are function wrapper. Do you know of any others?

If you don't know of others, does special casing this as the canonical output_wrapper seem like reasonable behavior to you?

FYI, if this were to become the default output_wrapper then the special case seems less surprising.

jart: Another example of an output wrapper would be `"%output%\n//# sourceMappingURL=foo_bin.js.map"`. But that's obviously not inside a function.

I would prefer not to special case this unless we have to. Would it be possible to solve this algorithmically? It almost reminds me of a popular interview question.

For posterity:

```
 --assume_function_wrapper              : Enable additional optimizations based
                                          on the assumption that the output
                                          will be wrapped with a function
                                          wrapper.  This flag is used to
                                          indicate that "global" declarations
                                          will not actually be global but
                                          instead isolated to the compilation
                                          unit. This enables additional
                                          optimizations.
```

jart: The interview question I'm thinking of, is the parentheses balancing one.

jart:
I think this contribution is good enough to merge as is. Want to rebase on master so the tests can run again? Then I'll hit merge.

hochhaus:
Rebased. I'm also happy to use the context sensitive paren matching to ensure that the output is indeed wrapped in a function. I just have been busy with higher priority items at the moment.

100% your call if you want to merge as-is or wait on a better implementation.

jart:
LGTM

Thank you!

jart:
Mr. Jenkins: test this please

PR #111, opened 2016-07-09
Implement correct runfiles behavior

jart:
This change is very important for being able to easily run a development web server in raw sources mode.

Assume your raw sources development web server is a `java_binary` rule. You obviously need to create a `closure_js_deps` file. Then you have to put that, along with all your source files, into the `data` attribute of your `java_binary`.

One might think to do that by having a zillion filegroups. This is pointless. Because the src/data graph is already defined by the closure rules. So you should only need one thing in the `data` attribute: a reference to the `closure_js_deps` rule.

This idealistic world of course assumes that everything is well defined. For example, a JS library should depend on CSS libraries whose classes it references. A CSS library should add a PNG file to its `data` attribute, if it references it. The same goes for template rules. Right now there isn't strict dependency checking for this stuff. But in the future, we'll write frontends for CSS/Soy compilers that strict dependency check images, CSS class names, and other data files.

This PR brings us one step closer to this brave new world of total website compilation.

The following behaviors have been implemented:

- data attribute on all rules
- js/css_library export no srcs
- js/css_library runfiles is srcs + data + transitive srcs/data
- js/css_binary runfiles is binary + data + transitive data - srcs
- js_deps runfiles is deps + data + transitive src/data

A bug has also been fixed where closure_js_deps wouldn't take into
consideration exported srcs.

CC: @hochhaus


closure/compiler/closure_js_library.bzl
```diff
@@ -83,15 +85,16 @@ def _impl(ctx):
...
                 js_language=determine_js_language(ctx),
                 js_exports=ctx.attr.exports,
                 js_provided=ctx.outputs.provided,
                 required_css_labels=collect_required_css_labels(ctx),
                 transitive_js_srcs=srcs,
                 transitive_js_externs=externs,
-                runfiles=ctx.runfiles(files=[ctx.outputs.provided,
-                                             ctx.outputs.stderr]))
+                transitive_data=tdata + ctx.files.data,
+                runfiles=ctx.runfiles(files=ctx.files.srcs + ctx.files.data,
```
mprobst: Shouldn't you be able to just use collect_default & co, and then wouldn't have to explicitly pass the transitive part?

mprobst:
Just one comment – I'm not sure I'm competent to review the whole PR, I don't know enough about how this works :-/

closure/compiler/closure_js_library.bzl
jart: Good question. I need to define the runfiles manually (rather than collect_{data,default}) because `closure_js_binary` implements a nonstandard behavior, which is to surgically remove the transitive sources it compiled from the runfiles.

But now that I think about it, we probably don't want to do this. Because even when the binary is compiled, the sourcemap still points to those files, which will need to be loaded when viewing a stacktrace in Chrome debugger.

I'll edit the PR to put them back shortly. That way I can use the collect_ params probably.

jart:
Mr. Jenkins: test this please

jart:
Mr. Jenkins: test this please

Issue #114, opened 2016-07-28
override warning_level


SalzzZ:
how to override default "--warning_level=VERBOSE"?
I tried adding

```
  defs = [
    "--warning_level=DEFAULT",
  ],
```

but it seems to be ignored.
Manually doing:

```
java -jar v20160713.jar --warning_level VERBOSE -- warning_level DEFAULT ...
```

output the js as expected

jart:
Unfortunately, there is currently no way to do this. Although it's something I've thought about in the past.

It would be very trivial though for us to change [closure_js_binary.bzl](https://github.com/bazelbuild/rules_closure/blob/master/closure/compiler/closure_js_binary.bzl) so it only sets the default value for that flag if one hasn't been provided in `defs`. I would happily merge something like that. Then you could use `git_repository()` to reference this repository until the next release.

jart:
Closing because referenced commit seems to fix this.

PR #120, opened 2016-08-03
add warning_level argument to closure_js_binary

SalzzZ:
the default will still be VERBOSE, but can be overridden by adding "warning_level" to closure_js_binary, see #114


bazel-io:

jart:
Thank you for taking the time to send us this change.

I'm LGTM'ing this change specifically because I want to help you be successful with Closure Rules.

I'd like for it to remain undocumented for the time being. At least until I can gain a better understanding of what the use cases for this new attribute are.

PR #131, opened 2016-09-09
Set //closure/testing/test:fontdemo to local

damienmg:
This genrule is broken by the sandbox because of a bug
in PhantomJS (ariya/phantomjs#13876).

Fixes bazelbuild/bazel#1715.

### DIFF
--- a/closure/testing/test/BUILD
+++ b/closure/testing/test/BUILD
@@ -85,6 +85,8 @@ genrule(
     outs = ["fontdemo-generated.png"],
     cmd = "$(location //third_party/phantomjs) $(SRCS) $@",
     tools = ["//third_party/phantomjs"],
+    local = 1,
+    tags = ["requires-network"],
 )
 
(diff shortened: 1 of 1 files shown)


damienmg:
To answer your question: it only applies to the genrule.

/cc @kchodorow

kchodorow:
LGTM

jart:
But isn't it only broken on Mac? Would it be possible for us to only disable this test on Mac?

damienmg:
Yes it is possible, do you prefer we do not run the test on mac?

damienmg:
(note that this change just make one action runs without sandboxing, the test itself is still sandboxed)

jart:
I thought the point of `local` was that it would only run on one's workstation. But I guess in the open source world it would still run on Travis. Glad I have it set up. So I guess this is ok to merge. I hope these bugs can be fixed in the future.

PR #132, opened 2016-09-09
Make contains_file() work across bazel versions

### LINKED ISSUE #130: contains_file() broken on bazel 0.3.0+
I noticed that the automatically populated `--defines` from `closure_js_binary()` are not being passed to the compiler after I upgraded my bazel version.

Please note that `is_using_closure_library` should be `True` in the following example.

```
ahochhaus@ahochhaus-pc:~/rules_closure$ git rev-parse HEAD
1eb589b4453bdb5b319f651bd614b05b753d0cd0
ahochhaus@ahochhaus-pc:~/rules_closure$ git diff
diff --git a/closure/compiler/closure_js_binary.bzl b/closure/compiler/closure_js_binary.bzl
index 41c7d48..86e54f1 100755
--- a/closure/compiler/closure_js_binary.bzl
+++ b/closure/compiler/closure_js_binary.bzl
@@ -43,6 +43,7 @@ def _impl(ctx):
   srcs, externs, tdata = collect_transitive_js_srcs(ctx)
   if not srcs:
     fail("There are no source files in the transitive closure")
+  print("is_using_closure_library: %s" % is_using_closure_library(srcs))
   _validate_css_graph(ctx)
   language_in = determine_js_language(ctx, normalize=True)
   language_out = ctx.attr.language
ahochhaus@ahochhaus-pc:~/rules_closure$ bazel build //closure/templates/test:greeter_test
WARNING: /home/ahochhaus/rules_closure/closure/compiler/closure_js_binary.bzl:46:3: is_using_closure_library: False.
INFO: Found 1 target...
Target //closure/templates/test:greeter_test up-to-date:
  bazel-bin/closure/templates/test/greeter_test
INFO: Elapsed time: 0.130s, Critical Path: 0.00s
ahochhaus@ahochhaus-pc:~/rules_closure$ bazel version
Build label: 0.3.1
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Jul 29 09:09:52 2016 (1469783392)
Build timestamp: 1469783392
Build timestamp as int: 1469783392
```

However `contains_file()` functions correctly on bazel 0.2.3. Please note that `is_using_closure_library` is `True` (as expected) in the example below.

```
ahochhaus@ahochhaus-pc:~/rules_closure$ bazel version
Build label: 0.2.3
Build target: bazel-out/local-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue May 17 14:21:13 2016 (1463494873)
Build timestamp: 1463494873
Build timestamp as int: 1463494873
ahochhaus@ahochhaus-pc:~/rules_closure$ bazel build //closure/templates/test:greeter_test
WARNING: /home/ahochhaus/rules_closure/closure/compiler/closure_js_binary.bzl:46:3: is_using_closure_library: True.
INFO: Found 1 target...
Target //closure/templates/test:greeter_test up-to-date:
  bazel-bin/closure/templates/test/greeter_test
INFO: Elapsed time: 66.174s, Critical Path: 35.26s
```

I believe the underlying cause is that File#short_path changed in bazel 0.3.0. https://github.com/bazelbuild/bazel/issues/1462

That API change also caused a bug in rules_go. https://github.com/bazelbuild/rules_go/issues/35

hochhaus:
Workaround of bazelbuild/bazel#1462

Fixed #130

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,6 @@ jdk:
 env:
   - V=HEAD URL=http://ci.bazel.io/job/Bazel/JAVA_VERSION=1.8,PLATFORM_NAME=linux-x86_64/lastSuccessfulBuild/artifact/output/ci/bazel--installer.sh FLAGS='--worker_verbose --strategy=Javac=worker --strategy=JsChecker=worker'
   - V=0.3.0 URL=http://bazel-mirror.storage.googleapis.com/github.com/bazelbuild/bazel/releases/download/0.3.0/bazel-0.3.0-installer-linux-x86_64.sh FLAGS=''
-  - V=0.2.2b URL=http://bazel-mirror.storage.googleapis.com/github.com/bazelbuild/bazel/releases/download/0.2.2b/bazel-0.2.2b-installer-linux-x86_64.sh FLAGS=''
 
 before_install:
   - wget -O install.sh $URL
--- a/closure/compiler/closure_js_binary.bzl
+++ b/closure/compiler/closure_js_binary.bzl
@@ -94,7 +94,7 @@ def _impl(ctx):
   if ctx.attr.pedantic:
     args += JS_PEDANTIC_ARGS
     args += ["--use_types_for_optimization"]
-  if contains_file(srcs, "external/closure_library/closure/goog/json/json.js"):
+  if contains_file(srcs, "../closure_library/closure/goog/json/json.js"):
     # TODO(hochhaus): Make unknownDefines an error for user supplied defines.
     # https://github.com/bazelbuild/rules_closure/issues/79
     args += ["--jscomp_off=unknownDefines",
--- a/closure/private/defs.bzl
+++ b/closure/private/defs.bzl
@@ -143,7 +143,7 @@ def determine_js_language(ctx, normalize=False):
   return language
 
 def is_using_closure_library(srcs):
-  return contains_file(srcs, "external/closure_library/closure/goog/base.js")
+  return contains_file(srcs, "../closure_library/closure/goog/base.js")
 
 def runpath(f):
   """Figures out the proper runfiles path for a file, using voodoo"""


bazel-io:

hochhaus:
@jart I'm not sure if this is the best solution or not. Does rules_closure care about maintaining compatibility with bazel 0.2.3 and earlier? If not, we can just update the hard coded paths. Let me know if you prefer that I go that route instead.

jart:
Travis currently tests against Bazel 0.2.2b. I'm willing to drop support for that if it makes things much faster and more elegant.

Can you give me some concrete examples to help me understand what's going on here?

Also consider the `runpath` function in `closure/private/defs.bzl`. We're already implementing workarounds with short path. Would your simpler solution make that go away too?

hochhaus:
Absolutely. From https://github.com/bazelbuild/rules_go/issues/35

> short_path of the file "@org_golang_x_sys//:unix/types_linux.go" used to be
> external/org_golang_x_sys/unix/types_linux.go but not it is ../org_golang_x_sys/unix/types_linux.go

In other words, in bazel 0.2.3 and before short_path evaluates to something along the lines of `external/org_golang_x_sys/unix/types_linux.go`. However in bazel 0.3.0 and later, short_path evaluates to something along the lines of `../org_golang_x_sys/unix/types_linux.go`. Note the `external/` vs `../` prefix.

This means that in `contains_file()` when we compare the short_path to a hard coded path which begins with `external/` that `False` is always returned in bazel 0.3.0+. This matters for cases like `is_using_closure_library()` and `--defines` detection.

jart:
Let's break 0.2.3 then. Would you mind updating this PR?

hochhaus:
Rebased and updated. PTAL.

jart:
Could you remove the line from our travis config that tests the old version of Bazel too? We don't want anyone to get the wrong idea about which versions are supported :)

hochhaus:
Removed. Thanks.

jart:
LGTM

PR #133, opened 2016-09-11
Set phantomjs_test to local

damienmg:
This will circumvent issue with Phantomjs failing in an OS X sandbox (ariya/phantomjs#13876)

All tests using phantomjs are currently failing. This just deactivate the sandbox for now
but might become a problem once we do remote execution in Bazel.

Tested on my mac laptop.


jart:
LGTM

PR #134, opened 2016-09-12
Avoid generated file naming conflict

hochhaus:
Allows both incremental_dom and traditional soy targets to
share the same soy template files without a naming conflict
in the generated JS files.


bazel-io:

jart:
LGTM

PR #136, opened 2016-09-20
Fix incremental-dom-closure.js generation for 0.5.0

hochhaus:
Add a target to ensure that missing required template @params result in a
compile time error. Without this bugfix (setting __esModule) in
incremental-dom-closure.js the compiler misses detection of required
parameters.

### DIFF
--- a/closure/templates/test/BUILD
+++ b/closure/templates/test/BUILD
@@ -16,6 +16,7 @@ package(default_testonly = True)
 
 licenses(["notice"])  # Apache 2.0
 
+load("//closure:defs.bzl", "closure_js_binary")
 load("//closure:defs.bzl", "closure_js_library")
 load("//closure:defs.bzl", "closure_js_proto_library")
 load("//closure:defs.bzl", "closure_js_template_library")
@@ -73,6 +74,17 @@ closure_js_library(
 )
 
 closure_js_library(
+    name = "greeter_idom_fail_lib",
+    srcs = ["greeter_idom_fail.js", "greeter_idom_fail_test.js"],
+    language = "ECMASCRIPT6_STRICT",
+    deps = [
+        ":greeter_idom_soy",
+        "//closure/library",
+        "//third_party/javascript/incremental_dom",
+    ],
+)
+
+closure_js_library(
     name = "greeter_proto_lib",
     srcs = ["greeter_proto.js"],
     deps = [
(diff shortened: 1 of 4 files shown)


bazel-io:

hochhaus:
The build failure is related to breaking changes in bazel HEAD and is not caused by this PR. Please see #137 for a fix of bazel at HEAD.

This change is read for review.

closure/templates/test/BUILD
```diff
@@ -116,6 +128,18 @@ closure_js_test(
...
+closure_js_binary(
+    name = "greeter_idom_fail_bin",
+    language = "ECMASCRIPT5_STRICT",
+    deps = [
+        ":greeter_idom_fail_lib",
+        "//closure/library",
+        "//closure/library:testing",
+    ],
+    pedantic = True,
+    internal_expect_failure = True,
```
jart: I absolutely love the fact that you're testing failure conditions. I had to put a lot of work into making that possible.

jart:
Please rebase this so we can test again.

closure/templates/test/BUILD
hochhaus: I was very happy when I saw it as an option! Outstanding work on the design to make it so simple.

One thing that I noticed was I needed to make the test a `closure_js_binary()` as `closure_js_test()` does not allow `internal_expect_failure`. On one hand this makes sense since the compile will fail so the test will (obviously) never be run. On the other hand it seems like a slightly asymmetrical API.

Can you share your thoughts here?

hochhaus:
Rebased. PTAL.

closure/templates/test/BUILD
jart: I would like to propagate the failure testing to all the other rules too. It's simply a question of time.

jart:
LGTM

Mr. Jenkins: test this please

closure/templates/test/BUILD
hochhaus: Thanks. As long as I know that is the intent I'll toss it on my list of projects that I can work on as time permits.

jart: Thanks for considering. It would be nice to know with greater certainty that the compile-time safety we're offering is working as advertised.

jart:
Now that all these contributions are in, I should be able to cut a release by tomorrow.

hochhaus:
> Now that all these contributions are in, I should be able to cut a release by tomorrow.

Thanks.

PR #143, opened 2016-09-30
Don't create a new JVM for each compilation

jart:
This commit causes `bazel build //closure/...` to go ~5x faster with RAM
consumption reduced by about 10GB. This was done by adding support for
Bazel worker processes to our modded version of the Closure Compiler.

To take advantage of this new development, you must put the following in
your ~/.bazelrc file:

    build --strategy=Closure=worker

This flag encompasses both JsChecker and JsCompiler. In the future, it
will also cover Closure Templates and Closure Stylesheets. However these
have not yet been incorporated into the worker process.

cc: @hochhaus


hochhaus:
I think this address the performance problem mentioned by @pcj in #77.

hochhaus:
LGTM

Thanks for implementing this.

pcj:
This sounds pretty boss.  Can't wait to try it out.

pcj:
Yes definitely faster, speedup was from ~9s (before) to ~3s (after) in the code I ran it against.  Awesome!

jart:
Glad you guys are digging the change!

I tried to incorporate Closure Templates compiler into this change. But private APIs and sealed jar prevented me from running that compiler in such a way that `System.exit()` wouldn't be called. I sent them some CLs internally to make those APIs public so we'll be able to speed up those compilations soon. But I think this commit represents major progress in the meantime.

Anyway I'm going to submit this now that the tests no longer flake. Thanks for taking a look guys.

pcj:
If this commit could talk...

big deal

PR #146, opened 2016-10-01
Add datatype check failure test

hochhaus:


bazel-io:

closure/compiler/test/BUILD
```diff
@@ -205,3 +205,26 @@ file_test(
...
+    srcs = ["type_fail.js"],
+    no_closure_library = True,
+)
+
+closure_js_binary(
+    name = "type_fail_bin",
+    internal_expect_failure = True,
+    pedantic = True,
+    deps = [":type_fail_lib"],
+    defs = ["--new_type_inf"],
```
hochhaus: This test does not fail unless I specify --new_type_inf. :(

hochhaus:
I believe that the failed test is actually just related to the breakage at HEAD.

closure/compiler/test/BUILD
jart: Traditional type inference doesn't recurse into literal data structures to ascertain type parameters. It only takes the raw type into consideration. Therefore it would fail to compile if you said:

```js
let /** !Array<string> */ foo = 123;
```

So when it comes to nested literals and traditional type inference, you have to help the compiler out by defining the type by hand. Just as if you were using Java. When using new type inference, don't even bother declaring the type because var/let/const essentially take on the same meaning as `auto` in C++.

But once Closure Compiler knows the correct parameterized type, it's good to go under either regime. For example, this program would fail just as you would expect.

```js
let /** !Array<string> */ foo = [123];
let /** !Array<number> */ bar = foo;
```

Oh you may have also encountered a bunch of errors that caused you to specify `no_closure_library`. That was my fault. It's fixed now.

hochhaus:
@jart Thanks for the detailed explanation. I never realized OTI worked like that.

PR #147, opened 2016-10-02
Fix travis bazel HEAD URL

hochhaus:


bazel-io:

jart:
Oh wow thank you so much for fixing this. I guess I can close out https://github.com/bazelbuild/continuous-integration/issues/32 now.

PR #148, opened 2016-10-02
Add unknown annotation failure test

hochhaus:


bazel-io:

closure/compiler/test/BUILD
```diff
@@ -205,3 +205,26 @@ file_test(
...
+    srcs = ["annotation_fail.js"],
+    no_closure_library = True,
+)
+
+closure_js_binary(
+    name = "annotation_fail_bin",
+    pedantic = True,
+    deps = [":annotation_fail_lib"],
+    # TODO(hochhaus): Enable the failure test
+    # internal_expect_failure = True,
```
hochhaus: This test should be failing, but is not

hochhaus:
This test is not a good reproducer of the behavior I was seeing in my code. Please hold off on review for the moment until I can update it.

PR #154, opened 2016-10-12
Use list for "--shouldGenerateGoogMsgDefs"

gramic:
Fixes: unsupported operand type(s) for +: 'list' and 'string'.

### DIFF
--- a/closure/templates/closure_js_template_library.bzl
+++ b/closure/templates/closure_js_template_library.bzl
@@ -49,7 +49,7 @@ def _impl(ctx):
     if ctx.attr.should_provide_require_soy_namespaces:
       args += ["--shouldProvideRequireSoyNamespaces"]
     if ctx.attr.should_generate_soy_msg_defs:
-      args += "--shouldGenerateGoogMsgDefs"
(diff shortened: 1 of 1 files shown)


bazel-io:

gramic:
I was too quick to request this pull.

An error states `If enabling shouldGenerateGoogMsgDefs, must also set bidi global directionality`.
In order to do that, I am planning to expose `bidiGlobalDir` in a way like the `orientation` of `closure_css_library` is implemented.

jart:
Thank you for sending this. I'd like to merge it anyway because that line of code is clearly incorrect. Even if this isn't the full solution.

If you could make bidi for templates work like I implemented it in the CSS rules, that would be great. Especially if you wrote a small regression test to make sure it keeps building.

But I must warn you in advance that the soy rules will probably be redesigned soon. The Closure Templates authors have come up with more modern skylark rules internally that are a significant improvement over the code I open sourced originally. Since this project is still technically in alpha, it might be nice if we open source the latest and greatest rules. If we go down that route, I'll make sure the upgrade process is as gentle as possible of course. I just want to give you fair warning if you're going to be investing time into improving these rules.

PR #159, opened 2016-10-28
Resolve correct artifact for HEAD and release candidate.

pcj:
Once ci.bazel.io put out a release candidate the initial osx travis script broke, demonstrating the importance of parsing the artifact name from the HTML page.  This change puts that logic back in (credit to @hochhaus) with pcregrep tweak for osx.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,8 @@ before_install:
   - CI_BASE="http://ci.bazel.io/job/Bazel/JAVA_VERSION=1.8,PLATFORM_NAME=$OS-$ARCH/lastSuccessfulBuild/artifact/output/ci"
   - CI_ARTIFACT="bazel--installer.sh"
   - URL="$GH_BASE/$GH_ARTIFACT"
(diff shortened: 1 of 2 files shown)


bazel-io:

jart:
Another nit: [optional] Consider using the `- |` syntax to have multiple lines in the shell script, so you don't need all the hyphens and one line if statements.

pcj:
Thanks, let me try out those suggestions.  I may not be able to get to this today.

hochhaus:
> Would it work if we just said grep -o 'bazel-.*-installer.sh'? I can't imagine that the perl syntax is necessary.

The non-greedy match was useful as `bazel-XXX-installer.sh` appears twice on the same line of HTML. The non-greedy match ensures avoids `bazel-` from matching the first instance and `-installer.sh` from matching the second instance and therefore all content in-between the two also being erroneously selected. The work around is trivial. This works on both macOS and linux for me.

```
$ curl -Ls $CI_BASE | grep -o 'bazel-[^-]*-installer.sh' | uniq
bazel--installer.sh
```

jart:
Wow thank you for looking into this. I've updated this in my local copy. It will be synced to GitHub in the next week or so.

PR #164, opened 2016-11-05
Update closure compiler and library to v20161024

hochhaus:


bazel-io:

jart:
LGTM

Mr. Jenkins: test this please

hochhaus:
Failure looks unrelated to this change. It is caused because the darwin installer is named `bazel-0.4.1-testrc1-installer.sh`. The non-greedy regexp should work here, but the `[^-]*` hack I proposed will fail since the version number itself has a `-` in it. We will need to find a better work around.

http://ci.bazel.io/job/Bazel/JAVA_VERSION=1.8,PLATFORM_NAME=darwin-x86_64/lastSuccessfulBuild/artifact/output/ci/

jart:
We're currently using the non-greedy pcregrep thing in the travis.yml file. I don't think I understand why it's not extracting that url correctly. But it is unrelated.

pcj:
I'm seeing that as well.  Works from command line though.  Apologies it's not perfect yet.

hochhaus:
I think we are using the default `CI_ARTIFACT="bazel--installer.sh"` when running on travis.

https://github.com/bazelbuild/rules_closure/blob/master/.travis.yml#L26

Correct me if I'm wrong, but on travis we run `if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then OS=darwin; fi`. This means that on travis OS will be darwin, not osx.

Then below we check $OS == osx. Shouldn't line 28 be the following (instead of osx)?

```
if [[ "$V" == "HEAD" && "$OS" == "darwin" ]];   then CI_ARTIFACT="`wget -qO- $CI_BASE | pcregrep -o 'bazel-.*?-installer.sh' | uniq`"; fi
```

jart:
Ohhh that makes sense. So it's a simple logic error.

jart:
https://github.com/bazelbuild/rules_closure/commit/037c09cc48455e46343f8f98a072040d8cacf5f0

PR #170, opened 2016-11-17
Fix Travis configuration

jart:


pcj:
Looks much cleaner, nice.  I'm still concerned this will fail if the installer has a different name however.  I've adopted this:

```
 - if [[ "$V" == "HEAD" ]]; then CI_ARTIFACT="`wget -qO- $CI_BASE | grep -o 'bazel-[-_a-zA-Z0-9\.]*-installer.sh' | uniq`"; fi
```

Reference:
https://github.com/pubref/rules_protobuf/blob/master/.travis.yml

jart:
If it has a different name for HEAD, I'll just update the configuration again, or put pressure on Bazel team to keep the name deterministic. The grep matching with that pattern is a little difficult to get right, as we've discovered.

pcj:
SGTM to keep the name deterministic.

jart:
Fixed in https://github.com/bazelbuild/rules_closure/commit/45b4f737b879cb489ff143d29b67b916eca5e76e

PR #174, opened 2016-12-02
Add missing suppress types

hochhaus:


bazel-io:

jart:
Wow thank you for finding all these. May I ask how you did it?

hochhaus:
Absolutely. My solution is a hack. It sounds like Tyler and the closure team have something much better in store.

The upshot is that I added a few bazel targets (see https://github.com/bazelbuild/rules_closure/pull/175) that generate a file which `goog.requires` every `goog.provided` symbol in the relevant library. Something similar to:

```
REQUIRE_ALL_COMMAND = "#/bin/bash\necho -e '/** @fileoverview Foo\n * @suppress {extraRequire}\n */' >$@; { for s in $(SRCS); do grep ^goog.provide\( $${s} || true; done; } | sed 's/goog.provide/goog.require/' | sort >>$@"
```

As the generated file only specifies `goog.requires` (with no `goog.provides`) it is included in the `LOOSE` build by default. Therefore, these generated files should cause the entire library to be compiled (and get full error checking).

The problem I faced was that the compiler only reports a subset of the errors on each invocation. Therefore, I needed to run the compiler in a loop adding one or more suppress at each invocation until the compile completed successfully (meaning the suppress clause was finally complete).

For this particular change I invoked the compiler with a hacky shell command but I can clear it up into something more general if you think it is helpful.

hochhaus:
Two other points that I missed.

My previous comment only talked about the non-generated libraries (ie: library, testing, jspb, soy_jssrc, idom). Finding the correct suppress types for the generated libraries is harder (eg: generated soy templates, generated protobufs). For any given invocation of a generated library it is difficult to guarantee that all code paths in the generator are executed. Therefore, I just manually found the missing suppress type for a small code base I have which is using `rules_closure`. I don't have a proposal to find a complete solution.

One other negative here is that it is difficult to locate extra suppress types (since many of these are not fully checked by JsChecker they don't have the "extra suppress" error check). I was thinking that for the non-generated libraries we could rebuild the suppress set starting from the empty set each time we updated a library. This should allow us to remove suppress types that are no longer used.

PR #175, opened 2016-12-02
Add rules to perform full compiles of libraries

hochhaus:
This is early in the development stages. @jart Can I have your feedback on the design as your time permits?

The general concept is to add rules (both for nti and oti) which compile each library in full (by generating a JS file that `goog.requires` every module in the corresponding libraries). This seems to work well. However,  the compiler only finds some of the errors during each pass. Therefore one needs to keep building these rules in a loop adding one or more additional suppress types per iteration. This loop terminates when the compile completes successfully (as as of the necessary suppress statement have been found).

Can you think of a better design? If not, I'll clean this up and submit something for a formal review.

### DIFF
--- a/closure/fullcompile/BUILD
+++ b/closure/fullcompile/BUILD
@@ -1,0 +1,119 @@
+# Copyright 2016 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


bazel-io:

MatrixFrog:
We're working on a similar/related project in the js compiler; we should coordinate. I'll try to take a look at this in the next few days.

hochhaus:
Thanks @MatrixFrog. Certainly this can be solved much more elegantly by the js compiler. I'm very interested to hear what the team is working on.

PR #182, opened 2017-01-25
Update proto rules to work with execroot rearrangement

kchodorow:
When https://github.com/bazelbuild/bazel/issues/1681 is submitted,
external repositories' proto sources will no longer be under the
execution root, they will be under ../repo_name. protoc needs to know
which directories to look under, so this updates the genrule to include
all repository paths in the protoc's search.

This change should be backwards and forwards compatible, since protoc
doesn't care if it gets extra search paths.


jart:
Thanks for sending this @kchodorow. Any chance you could mail this as a CL internally?

kchodorow:
Done.

PR #184, opened 2017-02-01
Update Java proto library and js_module_root refs to refer to sibling…

kchodorow:
… directories

This makes the js_module_root argument refer to the proper execution
root directory regardless of bazel version used.  Prep for https://github.com/bazelbuild/bazel/issues/1681.

Manually tested with old & new bazel versions.


jart:
Thank you for taking the time to update our code. I'm sorry to put you to the trouble, but would you mind patching this internally and sending me a CL? The internal repo is the source of truth. So I'd have to create a CL anyway if I merged this.

kchodorow:
Right!  Sorry, forgot about that.  Will do.

PR #186, opened 2017-02-10
Update closure-library to v20170124

pcj:
* Incorporates some relevant updates in `goog.net.streams.*`


bazel-io:

pcj:
On my end all tests pass except:

`//closure/testing/test:noto_fonts_render_as_expected                     FAILED in 0.0s`

with 

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
ERROR: FILES DO NOT HAVE EQUAL CONTENTS
closure/testing/test/fontdemo-darwin.png 0e48c707ba6f0054b946190eb8666fd60a7ae90a
closure/testing/test/fontdemo-generated.png c43598dc97762731af94a09c83829a77d9020445
```

Will wait for travis and jenkins to check it, suggestions welcome for fixing this failing test.

jart:
Thank you for this contribution. LGTM. I'm going to patch this PR internally. Then I'm going to push it back to GitHub. I'm also going to upgrade Closure Compiler and other tools for us.

The screenshot test being flaky on Mac is safe to ignore. We might even want to remove that test. The lord of web testing at Google, @DrMarcII, communicated to me one day that he's biased against screenshot testing. He's probably right.

hochhaus:
Thanks for the 0.4.1 release! I'm especially excited to use the new soy templates release (because it adds uniform support for the key attribute in idom). As always, rules_closure is going to make updating a breeze. :1st_place_medal:

jart:
My pleasure. Updating to 0.4.1 should be painless. Updates will continue to be painless, until it isn't.

Just a reminder for anyone reading: there will be a breaking release at some point in the future. We're just not going to tag it until we're 100% sure the pain will be worth it and there won't be much more to follow.

Also here's the latest news about this project, in case anyone is curious:

- TensorFlow has now [adopted](https://github.com/tensorflow/tensorflow/blob/master/WORKSPACE) Closure Rules (although at HEAD, which we still don't recommend for general consumption.)

- Thanks to the work @blickly has been doing on Closure Compiler, we're going to have a viable strategy for efficiently doing full authentic type checking within each `closure_js_library` rule. See: https://github.com/bazelbuild/rules_closure/commit/7412c0b3dabbdb1e07c72358bc790e44aafc313d

- We have a design doc in flight for a consolidated rule called `web_library` (currently known as webfiles in the codebase) and `web_binary` which is designed to unify the JS, CSS, and HTML compilers into a single multithreaded dæmon with AST caching and intelligent garbage collection. Many people from many orgs contributed to this design. It should make the [5x speedup](https://github.com/bazelbuild/rules_closure/pull/143) we saw earlier go to at least 10x. It will also have a development web server.

pcj:
Thanks for the update.  

The `web_binary` rule sounds interesting...  Can you explain a it more what it involves?  

Specifically, is the output of web_binary an artifact that is deployed to the browser (based on some newish technology that I have no knowledge of), or is only used for development, within rules_closure itself?

hochhaus:
Thanks @jart and @pcj!

I'm also interested in learning more about `web_binary`. Specifically can it easily be used in combination with another web server (ex: go's net/http) that implements application specific logic?

jart:
The output format is described by the following design doc: https://goo.gl/UGEiMn

The basic idea of webfiles is that each of its rules (web_library, web_binary, etc.) declares a set of assets, that maps Bazel paths to web server paths. In other words, each rule is its own self-contained website. The rule itself outputted to an efficient incremental protobuf format, which is an implementation detail. However each rule will also have an implicit NAME.zip output target which can be used for deploying it to a web server.

The tradeoff is that, if you use web_binary to compile a .js file, the compiled js file won't be part of the build graph, unless you use web_extract to pull it out of the set of web files and back into the build graph. This should be an acceptable trade-off. Usually people don't need to pull .class files out of the .jar file that java_library produces. So hopefully they shouldn't need to do the same here. But they can if they need to.

You'll absolutely be able to use it with Go's HTTP server. You can have your HTTP server rule depend on the zip and then extract it at startup. If you use this HTTP server for live source reloading development, then you'll most likely need to write a Skylark rule that integrates with the webfiles provider and read the protobufs directly for the `longpath` field. See the design doc. This is pretty straightforward. Go has good protobuf support.

hochhaus:
Thanks for the details.

> You'll absolutely be able to use it with Go's HTTP server.

Outstanding.

> You can have your HTTP server rule depend on the zip and then extract it at startup. If you use this HTTP server for live source reloading development, then you'll most likely need to write a Skylark rule that integrates with the webfiles provider and read the protobufs directly for the longpath field. See the design doc. This is pretty straightforward. Go has good protobuf support.

I'll happily implement this as time permits. Would it be useful to enough people to include this either here or in `rules_go`?

jart:
It's a Go dev web server, correct? If it's just a dev server, probably not. Dev servers tend to be highly tied to tools and architecture. For example, when writing [io.bazel.rules.closure.webfiles.server](https://github.com/bazelbuild/rules_closure/tree/master/java/io/bazel/rules/closure/webfiles/server) I had to rewrite a subset of @joeltine's internal JSDevServer, which is actually better, but was tied into an internal framework. Hopefully we'll be able to make that one open source at some point in the future. We just have a lot of things on the agenda.

But if your server is also able to help users get up and running deploying Closure Rules apps to production, possibly as a Go backend sitting behind something like NGINX, then I would consider merging it into a contrib/ folder. I'm sure many of our users would prefer to use Go instead of Java.

PR #198, opened 2017-04-27
Add basic support for bazel coverage

ribrdb:


bazel-io:

jart:
This is great. Thank you for this contribution. I'll patch this PR into our internal codebase and then re-export it to GitHub, with attribution, as soon as I get a chance.

ribrdb:
FYI I'm using this script to generate a report from the coverage data.
However I can't figure out any way to get bazel to run it over the test outputs.
https://gist.github.com/ribrdb/7969a336c333d0f62586e372cafec615

find bazel-testlogs/closure/testing/test/ -name cov.json|xargs python report.py

adam-rocska:
@jart && @dslomov : What about this PR nowadays? Will we lesser mortals also have coverage? 😁

adam-rocska:
@jart @ribrdb : Any chance this would get updated & merged? I would be more than happy to update the PR if you folks don't have the time, but it would be merged otherwise. I'd love to have these neat little coverage reports on my mortal little project 😃

closure/testing/phantomjs_harness.js
```diff
@@ -214,6 +225,7 @@ function onError(message, trace) {
     msg += t.file + ':' + t.line;
     system.stderr.writeLine(msg);
   });
+  writeCoverage();
```
gkdn: Not sure if it makes sense to write coverage on test failure.

closure/testing/phantomjs_harness.js
```diff
@@ -224,6 +236,7 @@ function onError(message, trace) {
  * @param {boolean} succeeded
  */
 function onCallback(succeeded) {
+  writeCoverage();
```
gkdn: Seems like Phantom can just report as coverage on this callback instead of creating another communication channel. If it is reported we can just write the results; we don't need another flag for phantom.

ribrdb:
Hmm, I haven't used this in ages. 
I can't remember the details but I had talked with someone on the bazel team about this and they had a couple problems with they way this was implemented. I think they wanted it to output using the lcov format instead of custom one, and that we also need to set instrumented_files:
https://docs.bazel.build/versions/master/skylark/rules.html#code-coverage-instrumentation
Although it looks like now they're planning to move to an xml coverage format:
https://docs.google.com/document/d/1-ZWHF-Q-qCKf19ik-t33ie58BkNurrYYzKR4OLtcilY/edit#

Maybe it's fine to go ahead with this though until the language-independent coverage support in bazel is finished.

dslomov:
@iirina any thoughts?

alexeagle:
FWIW I talked to @dslomov about JS coverage (in rules_nodejs) and we decided that since `bazel coverage` is currently in maintenance mode, it's fine to produce custom data and write it to `$TEST_UNDECLARED_OUTPUT_DIR` for some other tooling to pick up and process.

gkdn:
Thanks.

ribrdb:
No, I think it's better to not use closure compiler to run tests, and just use the coverage support in rules_nodejs.

PR #201, opened 2017-05-01
Bump closure library and closure compiler to recent versions

davido:
@args4j and @javax_inject are consumed now from the jar directory.
This probably can not stay that way and needs to be reverted.


bazel-io:

jart:
Thank you for your contribution. This is obsolete as of https://github.com/bazelbuild/rules_closure/commit/47a076d525d55f2e3bb3104d3f5ef4b60706f6af. (Although there's a build breakage I'm fixing right now.)

The `//jar` problem can be solved by using `java_import_external` for *everything* in your project. Cf. Nomulus codebase which is a great example of how to use Closure Rules. https://github.com/google/nomulus/blob/master/java/google/registry/repositories.bzl

PR #206, opened 2017-05-19
Add protoc compiler for Windows

meteorcloudy:
@jart Can you take a look, this will be needed by TensorFlow Bazel build on Windows.


jart:
Source of truth is internal.

I mailed you cl/156824366. Could you please check that it works on Windows?

PR #230, opened 2017-11-19
Use module resoltion NODE by default.

mrmeku:
Module resolution BROWSER appears to import a subset of the files supported by NODE
module resolution. The only difference being that when .js suffixes can be left off
of import statements and that non absolute/relative import statements will be imported
from a node_modules directory.


mrmeku:
Test this please.

PR #231, opened 2017-11-26
make @com_google_protobuf_java a proper external repository.

Ubehebe:
Currently, @com_google_protobuf_java is defined by the protobuf-java
jars published on Maven Central, imported into rules_closure by
the java_import_external() Skylark rule.

Meanwhile, the protobuf repository is a Bazel repository. Its
//:protobuf_java target provides the same functionality as the
protobuf-java jars. Importing the protobuf repo via http_archive
is the [officially recommended way of using protos in Bazel](https://blog.bazel.build/2017/02/27/protocol-buffers.html).

This commit changes the definition of rules_closure's
@com_google_protobuf_java from the maven jar-based approach to
a real http_archive. This will allow downstream repos that currently
have the recommended proto setup to use rules_closure.

### DIFF
--- a/closure/private/java_proto_library.bzl
+++ b/closure/private/java_proto_library.bzl
@@ -73,7 +73,7 @@ def java_proto_library(name, src, testonly=None, visibility=None, **kwargs):
   native.java_library(
       name = name,
       srcs = [":%s_srcjar" % name],
-      deps = ["@com_google_protobuf_java"],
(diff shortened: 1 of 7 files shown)


bazel-io:

Ubehebe:
BTW, I made this change with `buildozer 'replace deps @com_google_protobuf_java @com_google_protobuf_java//:protobuf_java' //...:*`, and checked it with `bazel build //...`.

PR #232, opened 2017-11-27
Use standard Bazel proto toolchain

Ubehebe:
This PR switches rules_closure to use Bazel's built-in java_proto_library rule instead of a custom Skylark rule. It also switches the protoc binary to the standard @com_google_protobuf//:protoc instead of a custom-downloaded dependency. See the commit descriptions for more details.

(Note that 73e9f1d is part of a [separate PR](https://github.com/bazelbuild/rules_closure/pull/231), so that diff should disappear from this PR once it is merged.)


bazel-io:

closure/repositories.bzl
```diff
@@ -681,79 +682,38 @@ def com_google_jsinterop_annotations():
       ],
   )
 
+def com_google_protobuf():
+  native.http_archive(
+    name = "com_google_protobuf",
+    sha256 = "0cc6607e2daa675101e9b7398a436f09167dffb8ca0489b0307ff7260498c13c",
+    strip_prefix = "protobuf-3.5.0",
+    urls = [
+      "https://mirror.bazel.build/github.com/google/protobuf/archive/v3.5.0.tar.gz",
```
Ubehebe: This bumps the protoc version from 3.3 to 3.5. (Because of some changes in the JSPB runtime, this requires a similar bump to com_google_protobuf_js below.) If you prefer, I can stick with 3.3 in this PR and bump the versions later.

jart:
Could you please send this as an internal CL? Otherwise I'll have to patch this by hand into the internal repo and re-export it.

PR #235, opened 2017-12-10
Support Java 9

davido:


bazel-io:

MatrixFrog:
Just as a reminder, the Closure Compiler itself can only contain Java 7 features, until further notice (but we're working on allowing Java 8 features and then Java 9 will hopefully follow soon after).

jart:
We also need to be compatible with App Engine which only supports Java 7.

davido:
> We also need to be compatible with App Engine which only supports Java 7.

Just to make sure, we avoid confusion. This change neither add Java 8 no Java 9 dependency to rules_closure. All it does, it upgrades transitive dependencies:

- auto-value
- dagger

It happens, that this upgrades allows to use rules_closure with Java 9. That's because, the upgraded versions use Generated annotation from Java 9, but only if it's available.

I verified, that this PR works as expected on Java 8!

jart:
Yes but are all these jars built as Java 7 binaries?

davido:
> Yes but are all these jars built as Java 7 binaries?

@jart Given your recent commit: fab2c6b1d5eeaf0b1a067054e425e8c38f667d40, that states: "After this change Closure Rules can no longer provide support for Java 7 runtime environments." Java 7 runtime compatibility shouldn't be an issue any more, right? Can we consider to merge this PR then?

bazel-io:

jart:
Does this change maintain ABI compatibility with Java 8?

davido:
> Does this change maintain ABI compatibility with Java 8?

Yes, it does. I investigated the approach from dependent projects (auto-common, auto-value and dagger), as I fixed the problem myself and uploaded PRs upstream and the solution was not to introduce any dependency on Java 9.

PR #238, opened 2017-12-23
Skylark compatability

### LINKED ISSUE #237: Bzl not forward compatible
`rules_closure` uses deprecated Skylark features. To reproduce, run this command with Bazel 0.9.0:

```
  $ bazel test --incompatible_depset_is_not_iterable=true javatests/...
..........
ERROR: /home/davido/projects/davido_rules_closure/java/io/bazel/rules/closure/BUILD:62:1: every rule of type _gensrcjar implicitly depends upon the target '@com_google_protobuf_protoc//:com_google_protobuf_protoc', but this target could not be found because of: no such package '@com_google_protobuf_protoc//': argument of type 'depset' is not iterable. in operator only works on lists, tuples, dicts and strings. Use --incompatible_depset_is_not_iterable=false to temporarily disable this check.
ERROR: Analysis of target '//javatests/com/google/javascript/jscomp:DiagnosticsTest' failed; build aborted: no such package '@com_google_protobuf_protoc//': argument of type 'depset' is not iterable. in operator only works on lists, tuples, dicts and strings. Use --incompatible_depset_is_not_iterable=false to temporarily disable this check
```

davido:
Closes #237.

This fixes "type 'depset' is not iterable. Use the `to_list()` method
to get a list." warning.

### DIFF
--- a/closure/filegroup_external.bzl
+++ b/closure/filegroup_external.bzl
@@ -20,8 +20,8 @@ def _filegroup_external(ctx):
   for different platforms.
   """
   strip_prefix = ctx.attr.strip_prefix
-  downloaded = depset()
-  basenames = depset()
+  downloaded = []
+  basenames = []
   data = depset(ctx.attr.data)
   inferred_srcs = depset()
   for sha256, urls, extract in _get_downloads(ctx):


bazel-io:

closure/filegroup_external.bzl
```diff
@@ -28,11 +28,11 @@ def _filegroup_external(ctx):
     basename = ""
     for url in urls:
       basename = url[url.rindex("/") + 1:] or basename
-      if url in downloaded:
+      if url in downloaded.to_list():
```
Yannic: Wouldn't it be better to define these as lists instead of casting them every time?

davido: Indeed. Done.

PR #239, opened 2017-12-23
Consume java_import_external.bzl from bazel

davido:


bazel-io:

Yannic:
Nice!

Submitted https://github.com/tensorflow/tensorflow/pull/33350 and https://github.com/tensorflow/tensorboard/pull/2773 to fix (some) consumers.

davido:
@Yannic Thanks for the clean up!

davido:
To provide some context for the reviewers:

@jart implemented first the `java_import_external` in rules_closure and re-used it in `tensorflow` and `tensorboard` projects. Later `java_import_external` rule implementation was upstreamed to Bazel, but unfortunately `rules_closure` and friends weren't adapted to use upstream version, unfortunately. So that this change is long overdue.

In particular current implementation of `java_import_external` conflicts with the incompatible option: [1] `--incompatible_load_java_rules_from_bzl` as pointed out by @ejona86 in https://github.com/bazelbuild/bazel/issues/8746#issuecomment-541246615.

The above option is going to be flipped in one of the next Bazel releases, so that sooner or later `rules_closure`  should switch to using upstream version (that hopefully will be fixed soon), instead of maintaining its own fork.

[1] https://github.com/bazelbuild/bazel/issues/8746

alexeagle:
Since you had to update tensorflow we can see this is a breaking change. However since the path has /private/ in it, I think this is a clear signal that this was never intended as part of the public API. Any idea if there's rules_closure documentation that guaranteed users they should rely on this in the public API?

We should make sure the next release includes release notes explaining how to make this change.

Yannic:
Well, this was exported in `defs.bzl`, so it was part of the public API. I think removing it is definitely the way to go, and the fix is pretty straight forward. The next release will already be a breaking change (see #423), so documenting it in the readme and the release note should be enough.

PR #255, opened 2018-02-23
add new bazel ci configuration

buchgr:
@jart Bazel is moving to a new CI system. We are checking in the configuration for our CI system so that we can test rules_closure after every Bazel change to ensure we don't break you. Cheers!


jart:
Thank you for maintenance. Could you send a corresponding CL to our third_party directory which is SoT?

buchgr:
sure. where should I send it to?

PR #261, opened 2018-03-16
Interoperate with rules_typescript

ribrdb:
There are two pieces to this:
 - An aspect which 'upgrades' a ts_library to a closure_js_library by running js_checker. Thus ts_libraries can be included in the deps of a closure_js_library or closure_js_binary. This part adds no dependencies to rules_closure. If users don't use rules_typescript the aspect will simply never do anything.
 - Support for generating .d.ts files for closure_js_libraries. This does add a dependency on clutz. However this seemed like the best way to implement this for several reasons:
     - This is the cleanest place to add the dependency, since clutz depends on the closure compiler.
     - It allows us to have a label for the generated .d.ts file, which wouldn't be possible if they were generated by an aspect.
     - The generated .d.ts file seems useful even to people who aren't using typescript. For example VS Code can read the .d.ts files to provide automatic imports and smart autocomplete in .js files.
     - I've tried to structure this so that clutz is not run (or even downloaded) unless the user is using typescript or manually builds the .d.ts files.

The corresponding changes to rules_typescript are at https://github.com/ribrdb/rules_typescript/tree/closure


jart:
Please see my email response on the rules_closure mailing list. We need the support of the rules_typescript to go forward with something like this.

jart:
@ribrdb 

@dslomov who will be leading the Closure Rules project going forward.

Please note we'd  @alexeagle because

LouisStAmour:
Just in case it helps anyone, here are two alternative approaches (read the full issue thread as I posted updated code later on) on integrating rules_typescript with rules_closure without code changes to either. Feedback welcome! https://github.com/bazelbuild/rules_typescript/issues/344

Issue #221, opened 2017-08-31
Error message on args4j version mismatch


davido:
I'm trying to upgrade ` args4j` from 2.0.26 to 2.33 in Gerrit Code Review project: [1]. For now, we consume the same ` args4j` versions as `rules_closure` and omitting fetching the `args4j` dependency in gerrit build.

After upgrading `args4j` to 2.33 in gerrit, rules_closue is failing with[2]:

```
No such file: io_bazel_rules_closure/closure/library:library
```

* [1] https://gerrit-review.googlesource.com/67793
* [2] http://paste.openstack.org/show/620140

jart:
I tried upgrading this dependency for you. Here's what happened:

```
ERROR: /usr/local/google/home/jart/code/rules_closure/java/io/bazel/rules/closure/webfiles/server/BUILD:57:1: Generating SOY v2 Java files //java/io/bazel/rules/closure/webfiles/server:listing_files failed (Exit 1)
INTERNAL SOY ERROR.
Please open an issue at https://github.com/google/closure-templates/issues with this stack trace and repro steps
java.lang.IllegalStateException: Cannot set value to a final field 'com.google.template.soy.AbstractSoyCompiler.protoFileDescriptors'.
        at org.kohsuke.args4j.spi.Setters.create(Setters.java:32)
        at org.kohsuke.args4j.ClassParser.parse(ClassParser.java:34)
        at org.kohsuke.args4j.CmdLineParser.<init>(CmdLineParser.java:96)
        at org.kohsuke.args4j.CmdLineParser.<init>(CmdLineParser.java:71)
        at com.google.template.soy.SoyCmdLineParser.<init>(SoyCmdLineParser.java:43)
        at com.google.template.soy.AbstractSoyCompiler.doMain(AbstractSoyCompiler.java:204)
        at com.google.template.soy.AbstractSoyCompiler.run(AbstractSoyCompiler.java:183)
        at com.google.template.soy.AbstractSoyCompiler.runMain(AbstractSoyCompiler.java:175)
        at com.google.template.soy.SoyParseInfoGenerator.main(SoyParseInfoGenerator.java:105)
```

I encourage you to file an issue with https://github.com/google/closure-templates to help their codebase be compatible with the newer version.

davido:
> I encourage you to file an issue with https://github.com/google/closure-templates to help their codebase be compatible with the newer version.

It's all addressed any fixed already: [1].

[...]

PR #262, opened 2018-03-18
Bump args4j to 2.33

davido:
Fixes: #221.

closure-template version is also updated to include this PR: [1].

[1] https://github.com/google/closure-templates/pull/155


jart:
Thank you for the contribution. This will be addressed by a shortly upcoming commit.

jart:
Out of curiosity, why does args4j need to be updated? com.google.template:soy:2018-03-14 does not declare a dependency on this newer version. Google internally still uses 2.0.26.

Issue #190, opened 2017-03-18
Proto libraries with dependencies


wiktortomczak:
Is there a way to generate a JS proto library from a .proto file with dependencies (on other .proto files that the top-level .proto imports) ? IIUC [closure_js_proto_library](https://github.com/bazelbuild/rules_closure#closure_js_proto_library) does not support a deps= argument. I suppose it could?

I believe the implementation of closure_js_proto_library(deps=...) could rely on the [proto_library](https://bazel.build/blog/2017/02/27/protocol-buffers.html) rule added to Bazel recently.

rohanag12:
Is there any plan on supporting this feature? Also, is there some workaround we can use for this (other than bundling statically generated files, of course)?

shubhangitaneja:
Ability to specify deps in closure_js_proto_library would be definitely helpful. My proto is importing "google/protobuf/timestamp.proto" and right now I don't see any way to specify a dependency on it.

pshields:
I've run into a similar issue. The protos I'm compiling import `google/api/annotations.proto`. I tried adding that file and its deps as a filegroup passed to `srcs` (in addition to my regular `srcs`), but ran into additional errors. I think being able to specify deps via a `deps` parameter would help. At the very least, it would be more idiomatic.

Incidentally, if I strip out my  `google/api/annotations.proto` import and related code, my `closure_js_proto_library` rule is able to build successfully and is able to resolve my `google/protobuf/timestamp.proto` imports.

PR #263, opened 2018-04-12
Modernize protobuf rule

Yannic:
This CL adds a new proto rule which works with Bazel's native |proto_library| rules.
We also upgrade travis to a newer version of Bazel.

Closes bazelbuild/rules_closure#190

cc: @hochhaus


jameswex:
@Yannic I just started some work in tensorboard and needed this exact thing to exist, so thanks!

@jart Patching in this PR, then updating tensorboard's 3 collect_js calls to match the now-changed API, I was able to build the example protos in tensorflow using closure_proto_library and have that as a dep in my tensorboard tf_web_library component and successfully see it work in my tensorflow_html_binary demo server for that new component.

jart:
verified test this please (I hope this is the right incantation for Bazel CI these days)

.travis.yml
```diff
@@ -33,22 +33,28 @@ before_install:
 
 script:
   - |
+    excludes=()
+    if [[ "${TRAVIS_OS_NAME}" == "osx" ]]; then
+      # https://github.com/bazelbuild/rules_closure/issues/247
+      excludes+=('//closure/testing/test:noto_fonts_render_as_expected')
```
jart: I would welcome a PR commenting out that test rule, saying, "hey this sometimes works, and is an example of what you *could* do, depending on your environment." If you do this, please also revert the complexity that rule added to this file.

Yannic:
@jart it's not
https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#pull-requests

closure/compiler/closure_js_library.bzl
```diff
@@ -33,28 +33,49 @@ load("//closure/private:defs.bzl",
...
+  if file:
+    return file
+  return actions.declare_file(name)
 
-  # Create a list of the sources defined by this specific rule.
-  srcs = ctx.files.srcs
-  if ctx.files.externs:
-    print("closure_js_library 'externs' is deprecated; just use 'srcs'")
-    srcs = ctx.files.externs + srcs
+def closure_js_library_impl(
```
jart: There's a lot of surface area change here. I feel comfortable with this because you took the time to write more tests. Thank you. I'll do further integration testing with dependent projects afterwards, at some point.

closure/compiler/closure_js_library.bzl
```diff
@@ -33,28 +33,49 @@ load("//closure/private:defs.bzl",
...
+    actions, label, workspace_name,
+
+    srcs, deps, testonly, suppress,
+    closure_library_base, closure_library_deps, _ClosureWorker,
+
+    includes=[],
```
jart: Please note that Skylark may have the same problem as Python. If you say `includes=[]` instead of `includes=()` then code might accidentally mutate the default value. That doesn't seem to be happening here.

closure/protobuf/closure_proto_library.bzl
```diff
@@ -0,0 +1,167 @@
...
+"""
+
+load("//closure/compiler:closure_js_library.bzl", "closure_js_library_impl")
+load("//closure/private:defs.bzl",
+     "CLOSURE_WORKER_ATTR",
+     "CLOSURE_LIBRARY_BASE_ATTR",
+     "CLOSURE_LIBRARY_DEPS_ATTR",
+     "unfurl")
+
+# This was taken from https://github.com/bazelbuild/rules_go/blob/67f44035d84a352cffb9465159e199066ecb814c/proto/compiler.bzl#L72
```
jart: We don't take things. We're just borrowing code with the same copyright / license :)

jart:
Thanks for the contribution. I'm merging now to get folks unblocked. LGTM

PR #273, opened 2018-06-05
Remove unused filegroup_external

davido:
Last usage of filegroup_external was removed in d3e0437, but it wasn't
removed in that change. Do it now and clean it up.


jart:
Thanks for taking the time to send this PR but I <3 `filegroup_external` https://github.com/tensorflow/tensorboard/blob/master/third_party/js.bzl#L25

PR #277, opened 2018-06-23
closure_proto_library: propagate descriptors to closure_js_library

pcj:
Makes closure_proto_library a valid dependency for
closure_js_template_library rules.

Update test to use modern proto rule.

### DIFF
--- a/closure/protobuf/closure_proto_library.bzl
+++ b/closure/protobuf/closure_proto_library.bzl
@@ -93,6 +93,10 @@ def _closure_proto_aspect_impl(target, ctx):
   srcs = depset([js])
   deps = unfurl(ctx.rule.attr.deps, provider="closure_js_library")
   deps += [ctx.attr._closure_library, ctx.attr._closure_protobuf_jspb]
+  proto_deps = unfurl(ctx.rule.attr.deps, provider="proto")
+  descriptors = depset(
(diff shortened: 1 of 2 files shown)


closure/protobuf/closure_proto_library.bzl
```diff
@@ -93,6 +93,10 @@ def _closure_proto_aspect_impl(target, ctx):
   srcs = depset([js])
   deps = unfurl(ctx.rule.attr.deps, provider="closure_js_library")
   deps += [ctx.attr._closure_library, ctx.attr._closure_protobuf_jspb]
+  proto_deps = unfurl(ctx.rule.attr.deps, provider="proto")
+  descriptors = depset(
+    direct = target.files.to_list(),
+    transitive=[d.proto.transitive_descriptor_sets for d in proto_deps])
```
Yannic: I think it should be sufficient to skip the `unfurl` and use
```
descriptors = depset(
  direct = target.proto.direct_descriptor_set,
  transitive = target.proto.transitive_descriptor_sets)
```

pcj: Thanks!  Added a commit to address this (both arguments want a sequence though).

pcj:
@dslomov @jart Can we get this merged?  Cannot use protobuf with closure templates without this.

PR #278, opened 2018-08-04
Export closure_proto_aspect and document closure_grpc_web_library

Yannic:
Exposing the aspect is required for the implementation of
closure_grpc_web_library in https://github.com/grpc/grpc-web/


v48:
This sounds awesome! I would like to try `closure_grpc_web_library` for my project Is there a possibility to merge this soon? Im currently generating files with a separate script it would be awesome to use bazel instead.

v48:
@Yannic Would it be possible to get this change in soon? I would really like to try this bazel rule in my project.

v48:
@dslomov Thanks for rebasing and updating 🎉 , is there an eta on when this will be released?

PR #282, opened 2018-08-22
Fix build on JDK9

davido:
Fixes: https://github.com/bazelbuild/bazel/issues/5961


laurentlb:
@dslomov

davido:
It appears not to be needed any more, as it was fixed in latest Bazel version.

PR #314, opened 2018-11-26
closure_proto_library: propagate descriptors to create_closure_js_library (like #277)

factuno-db:
This makes PR #277 mergeable.


closure/protobuf/closure_proto_library.bzl
```diff
@@ -90,13 +90,16 @@ def _closure_proto_aspect_impl(target, ctx):
...
+        direct = [target.proto.direct_descriptor_set],
+        transitive = [target.proto.transitive_descriptor_sets])
 
     suppress = [
         "missingProperties",
         "unusedLocalVariables",
     ]
 
-    library = create_closure_js_library(ctx, srcs, deps, [], suppress, True)
+    library = create_closure_js_library(ctx, srcs, deps, [], suppress, True, descriptors)
```
Yannic: Could we change this to keyword arguments? This starts getting hard to parse...
```python
library = create_closure_js_library(
    ctx = ctx,
    srcs = srcs,
    deps = deps,
    exports = [],
    suppess = suppress,
    lenient = True,
    internal_descriptors = descriptors,
)
```

closure/compiler/closure_js_library.bzl
```diff
@@ -45,7 +45,8 @@ def create_closure_js_library(
         deps = [],
         exports = [],
         suppress = [],
-        lenient = False):
+        lenient = False,
+        internal_descriptors = []):
```
Yannic: Could we use a more descriptive name than `internal_descriptors`? 
(I know that it is called that in `_closure_js_library_impl`, but this is public and might be harder to change later)

closure/templates/test/BUILD
```diff
@@ -18,17 +18,22 @@ licenses(["notice"])  # Apache 2.0
 
 load("//closure:defs.bzl", "closure_js_binary")
```
Yannic: Please copy the existing tests to `closure/templates/test/legacy` so we don't lose coverage here. I'm not sure if we can remove `closure_js_library` anytime soon...

factuno-db:
Thanks for the review! Should be fixed with the latest commits.

closure/protobuf/closure_proto_library.bzl
```diff
@@ -90,13 +90,23 @@ def _closure_proto_aspect_impl(target, ctx):
     srcs = depset([js])
     deps = unfurl(ctx.rule.attr.deps, provider = "closure_js_library")
     deps += [ctx.attr._closure_library, ctx.attr._closure_protobuf_jspb]
+    descriptors = depset(
+        direct = [target.proto.direct_descriptor_set],
+        transitive = [target.proto.transitive_descriptor_sets])
```
Yannic: nit: Please add a trailing `,` and move `)` to the next line.

closure/protobuf/closure_proto_library.bzl
```diff
@@ -90,13 +90,23 @@ def _closure_proto_aspect_impl(target, ctx):
...
 
-    library = create_closure_js_library(ctx, srcs, deps, [], suppress, True)
+    library = create_closure_js_library(
+        ctx = ctx,
+        srcs = srcs,
+        deps = deps,
+        exports = [],
+        suppress = suppress,
+        lenient = True,
+        pb_descriptors = descriptors)
```
Yannic: nit: Dito

gkdn:
I didn't dig down the PR deeply but making closure_js_library understand from PB doesn't seem right; plus making it part of its public API.

gkdn:
@factuno-db I'm looking at commit 67c1337 but cannot see how this might change the behavior. Could you clarify why this breaks type checking?

factuno-db:
> @factuno-db I'm looking at commit [67c1337](https://github.com/bazelbuild/rules_closure/commit/67c1337aa6348a395e47b5a100ceb412b2c69bd6) but cannot see how this might change the behavior. Could you clarify why this breaks type checking?

I'm sorry, I have mistaken the commit. That was the commit that made #277 unmergeable.

gkdn:
@factuno-db I looked at original commit.
The problem is general approach doesn't seem correct. Looking at the code proto rule; descriptors are communicated through closure library provider. However instead there should be an extra provider for the proto that should provide the information necessary (i.e. descriptors)

factuno-db:
Alright. I will close this pull request and try to implement it via an extra provider. Thanks a lot!

PR #313, opened 2018-11-24
Bump protobuf version to 3.6.1.2

davido:

PR #318, opened 2018-12-06
Bump protobuf version to 3.6.1.2 (2)

meteorcloudy:
Following up https://github.com/bazelbuild/rules_closure/pull/313/files, where we missed `com_google_protobuf_js`

/cc @davido @buchgr @laurentlb

### DIFF
--- a/closure/repositories.bzl
+++ b/closure/repositories.bzl
@@ -693,7 +693,7 @@ def com_google_protobuf_js():
             "https://github.com/protocolbuffers/protobuf/archive/v3.6.1.2.tar.gz",
         ],
         sha256 = "2244b0308846bb22b4ff0bcc675e99290ff9f1115553ae9671eba1030af31bc0",
-        strip_prefix = "protobuf-7b28271a61a3da0a37f6fda399b0c4c86464e5b3/js",
+        strip_prefix = "protobuf-3.6.1.2/js",
         build_file = str(Label("//closure/protobuf:protobuf_js.BUILD")),
     )
 
--- a/closure/repositories.bzl
+++ b/closure/repositories.bzl
@@ -680,6 +680,7 @@ def com_google_protobuf():
         strip_prefix = "protobuf-3.6.1.2",
         sha256 = "2244b0308846bb22b4ff0bcc675e99290ff9f1115553ae9671eba1030af31bc0",
         urls = [
+            "https://mirror.bazel.build/github.com/google/protobuf/archive/v3.6.1.2.tar.gz",
             "https://github.com/protocolbuffers/protobuf/archive/v3.6.1.2.tar.gz",
         ],
     )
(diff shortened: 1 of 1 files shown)


buchgr:
LGTM

meteorcloudy:
@gkdn I am seeing this error, any idea?
```
external/com_google_protobuf_js/binary/utils.js:974: ERROR - variable Buffer is undeclared
  if (data.constructor === Buffer) {
                           ^
  ProTip: "JSC_UNDEFINED_VARIABLE" or "checkVars" or "missingSourcesWarnings" or "undefinedVars" can be added to the `suppress` attribute of:
  //closure/protobuf:jspb
  Alternatively /** @suppress {undefinedVars} */ can be added to the source file.
```

davido:
@meteorcloudy thanks for the fix.

davido:
@meteorcloudy I can reproduce the breakage and I have an idea where it's coming from.

Previously, protobuf was v3.5.1, we have upgraded it to v3.6.1.2.
The root case for the breakage, is the `js/binary/utils.js` file and the fact that Buffer class is only known in Node, and closure compiler knows nothing about, thus the breakage.

The better question is, why it worked in v3.5.1 and stopped working in v3.6.1.2 (this file is identical in v3.6.1, so it wouldn't pass rulese_clsoure tests as well).

Here is the difference between v3.5.1 and v3.6.1.2:

```diff
diff --git a/utils.js b/var/tmp/utils_v3.6.1.js
index c706bff4..55a9ccd4 100644
--- a/utils.js
+++ b/var/tmp/utils_v3.6.1.js
@@ -431,7 +431,7 @@ jspb.utils.joinHash64 = function(bitsLow, bitsHigh) {
 
 /**
  * Individual digits for number->string conversion.
- * @const {!Array.<string>}
+ * @const {!Array<string>}
  */
 jspb.utils.DIGITS = [
   '0', '1', '2', '3', '4', '5', '6', '7',
@@ -554,10 +554,10 @@ jspb.utils.hash64ToDecimalString = function(hash, signed) {
 /**
  * Converts an array of 8-character hash strings into their decimal
  * representations.
- * @param {!Array.<string>} hashes The array of hash strings to convert.
+ * @param {!Array<string>} hashes The array of hash strings to convert.
  * @param {boolean} signed True if we should treat the hash string as encoding
  *     a signed integer.
- * @return {!Array.<string>}
+ * @return {!Array<string>}
  */
 jspb.utils.hash64ArrayToDecimalStrings = function(hashes, signed) {
   var result = new Array(hashes.length);
@@ -971,8 +971,12 @@ jspb.utils.byteSourceToUint8Array = function(data) {
     return /** @type {!Uint8Array} */(new Uint8Array(data));
   }
 
+  if (data.constructor === Buffer) {
+    return /** @type {!Uint8Array} */(new Uint8Array(data));
+  }
+
   if (data.constructor === Array) {
-    data = /** @type {!Array.<number>} */(data);
+    data = /** @type {!Array<number>} */(data);
     return /** @type {!Uint8Array} */(new Uint8Array(data));
   }

```

Git blame reveals that this commit is missing:

https://github.com/protocolbuffers/protobuf/commit/f00e06c95bc117fb2ed0ca56c96041c93039f1fe

davido:
@meteorcloudy See my two suggestions how to rectify this breakage in https://github.com/protocolbuffers/protobuf/pull/4650#issuecomment-445002198.

davido:
@meteorcloudy I uploaded this PR to fix the test breakage: https://github.com/bazelbuild/rules_closure/pull/319. With this patch, all tests are passing. Of course it would be cleaner if we would have 3.6.1.3 protobuf release that addresses that problem.

PR #326, opened 2019-01-18
Run buildifier --lint=fix

laurentlb:
This fixes multiple issues, related to future changes in Bazel.

Remaining work: make the code compatible with `--incompatible_depset_is_not_iterable` and `--incompatible_depset_union`.


closure/filegroup_external.bzl
```diff
@@ -28,11 +28,11 @@ def _filegroup_external(ctx):
         basename = ""
         for url in urls:
             basename = url[url.rindex("/") + 1:] or basename
-            if url in downloaded:
+            if url in downloaded.to_list():
```
gkdn: Isn't there a more efficient way to do contains check in depssets?

closure/private/defs.bzl
```diff
@@ -188,7 +188,7 @@ def find_js_module_roots(srcs, workspace_name, label, includes):
                         break
                 if not found:
                     fail("No srcs found beginning with '%s/'" % include)
-            for root in roots:
+            for root in roots.to_list():
```
gkdn: is to_list cached if called multiple times on a depset?

laurentlb:
The rule isn't using depsets in an optimal way. This is why it doesn't work with `--incompatible_depset_is_not_iterable` and `--incompatible_depset_union` - and we should fix that.

The calls to `to_list` were inserted automatically by `buildifier --lint=fix`. It simply makes more explicit what was happening behind the scene.

I can either leave the code like this, or revert the two lines you pointed out. In both cases, this should ideally be revisited in the future.

laurentlb:
Please merge it (I don't have write access)

PR #332, opened 2019-02-23
Move default Closure compiler dependency to v20190215

nreid260:


gkdn:
(Closing since closure dependency is already updated.)

PR #337, opened 2019-03-13
Fix phantom-js for linux environment

gkdn:


gkdn:
@dslomov Currently linux builds are broken. This fixes the issue locally for me and for travis as well. buildkite failed for another reason but I cannot re-trigger the build.

PR #339, opened 2019-03-13
Allow dict for entry_points.

### LINKED ISSUE #323: closure_js_test does not allow multiple entry_points
When running a closure_js_test rule an entry point needs to be provided to not have the compiler remove all code. 

I have a couple of tests which are goog.modules with different module names. The closure_js_test currently only allows a single entry point. When adding more than a single entry point the test fails.

So currently there needs to be a closure_js_test rule for each test with a different module name.

michaelfig:
Fixes #323

This feature allows writing a single `closure_js_test` rule when each source file's entry_points are different.

For example:

```python
closure_js_test(
    name = "test_lib",
    srcs = glob([
        "*_test.js",
    ]),
    entry_points = {
        "binary_keyset_reader_test.js": ["tink.BinaryKeysetReaderTest"],
        "cleartext_keyset_handle_test.js": ["tink.CleartextKeysetHandleTest"],
        "config_test.js": ["tink.ConfigTest"],
        "crypto_format_test.js": ["tink.CryptoFormatTest"],
        "keyset_handle_test.js": ["tink.KeysetHandleTest"],
        "primitive_set_test.js": ["tink.PrimitiveSetTest"],
        "proto_test.js": ["tink.ProtoTest"],
        "registry_test.js": ["tink.RegistryTest"],
        "util_test.js": ["tink.UtilTest"],
    },
[...]
```

This compresses the build rules significantly.

Comments or questions are appreciated,
Michael.


gkdn:
Patch seems ok but needs some tests.

BTW, I think if you change it so that you can pass dependency_mode=SORT_ONLY, I think you can get away without even providing entry_points.

michaelfig:
Thanks for the review!  I will test with dependency_mode and update accordingly with tests.

michaelfig:
Hi @gkdn, sorry for the long delay.

`SORT_ONLY` is not supported by the closure compiler used by rules_closure.  I can pass `dependency_mode=LOOSE`, but that doesn't seem to make any difference.

For a test where I have `aes_gcm_key_manager_test.js`:

```javascript
goog.module("tink.aead.AesGcmKeyManagerTest");
...
```

And the following `BUILD.bazel` rule:

```
closure_js_test(
    name = "test_lib",
    testonly = 1,
    lenient = True,
    srcs = glob([
        "*_test.js",
    ]),
    dependency_mode = "LOOSE",
    deps = [
    ],
...)
```

The generated `test_lib_aes_gcm_key_manager_bin.js` is empty, and the library dependency, `test_lib_aes_gcm_key_manager_lib.i.js` has the contents of `aes_gcm_key_manager_test.js` in it, but it is not referenced in the final closure-compiled test.

It only all links together when I supply `entry_points = ["tink.aead.AesGcmKeyManagerTest"]` to the `closure_js_binary` rule.

Do you have any other advice?

gkdn:
`SORT_ONLY` is supported by the closure compiler. See an example here:
https://github.com/google/j2cl/blob/48f1ea6ae847277989da0fce5c36117074e40833/jre/javatests/BUILD#L162

`closure_js_test` doesn't seem to  have a `dependency_mode` attribute, but you should be able to use `defs` to pass it.

BTW, I suggested SORT_ONLY just as an alternative quick route for you; otherwise your patch makes sense; just only missing a simple test.

michaelfig:
Unfortunately, `--dependency_mode=SORT_ONLY` reports errors in the closure library, so I really do need `entry_points`.  Thanks for the suggestion, though.

FWIW, with:
```
closure_js_test(
    name = "goog_testSuite_test",
    srcs = ["goog_testSuite_test.js"],
    defs = [
        "--dependency_mode=SORT_ONLY",
    ],
    deps = [
        ":goog_testSuite_lib",
        "//closure/library:testing",
    ],
)
```

I get:

```
ERROR: /Users/michael/src/self/rules_closure/closure/testing/test/BUILD:107:1: Compiling 275 Ja
vaScript files to closure/testing/test/goog_testSuite_test_bin.js failed (Exit 1)
external/com_google_javascript_closure_library/closure/goog/crypt/hashtester.js:199: ERROR - Va
lue assigned to local variable digest is never read
        var digest = hash.digest();
            ^
  ProTip: "JSC_UNUSED_LOCAL_ASSIGNMENT" or "unusedLocalVariables" can be added to the `suppress` attribute of:
  //closure/library/crypt:hashtester
  Alternatively /** @suppress {unusedLocalVariables} */ can be added to the source file.

external/com_google_javascript_closure_library/closure/goog/storage/mechanism/mechanismtestdefinition.js:26: ERROR - Variable mechanism declared more than once. First occurrence: external/com_google_javascript_closure_library/closure/goog/storage/mechanism/iterablemechanismtester.js
var mechanism;
    ^
  ProTip: "JSC_VAR_MULTIPLY_DECLARED_ERROR" or "checkVars" or "duplicate" or "missingSourcesWarnings" can be added to the `suppress` attribute of:
  //closure/library/storage/mechanism:mechanismtestdefinition
  Alternatively /** @suppress {duplicate} */ can be added to the source file.

external/com_google_javascript_closure_library/closure/goog/storage/mechanism/mechanismtester.js:33: ERROR - Variable mechanism declared more than once. First occurrence: external/com_google_javascript_closure_library/closure/goog/storage/mechanism/iterablemechanismtester.js
var mechanism = null;
    ^
  ProTip: "JSC_VAR_MULTIPLY_DECLARED_ERROR" or "checkVars" or "duplicate" or "missingSourcesWarnings" can be added to the `suppress` attribute of:
  //closure/library/storage/mechanism:mechanismtester
  Alternatively /** @suppress {duplicate} */ can be added to the source file.

3 error(s), 0 warning(s)

Target //closure/testing/test:goog_testSuite_test failed to build
```

closure/testing/test/BUILD
```diff
@@ -83,6 +83,22 @@ closure_js_test(
...
 
+closure_js_library(
+    name = "goog_testSuite_lib",
+    srcs = ["goog_testSuite.js"],
+)
+
+closure_js_test(
+    name = "goog_testSuite_test",
+    srcs = ["goog_testSuite_test.js"],
+    entry_points = {
```
gkdn: To double check, does your test fail when we remove entry_points?

closure/testing/test/BUILD
```diff
@@ -83,6 +83,22 @@ closure_js_test(
     ],
 )
 
+closure_js_library(
+    name = "goog_testSuite_lib",
+    srcs = ["goog_testSuite.js"],
+)
```
gkdn: Seems like you don't need the library to reproduce your issue.

closure/testing/test/BUILD
```diff
@@ -83,6 +83,22 @@ closure_js_test(
...
+
+closure_js_test(
+    name = "goog_testSuite_test",
```
gkdn: name = "dict_entry_point_test"

closure/testing/test/BUILD
michaelfig: Yes, I've verified the failure without the specific `entry_points` entry.

closure/testing/test/BUILD
michaelfig: You're right!  I've removed the library.

closure/testing/test/BUILD
michaelfig: Changed.

PR #340, opened 2019-03-13
Allow lenient and empty srcs to typecheck

michaelfig:
These changes allow `closure_js_library` to pass when only `exports` are supplied, as well as passing `closure_js_test` when `lenient = True` is supplied.


gkdn:
IIUC, you would like to solve two problems;
1. empty srcs to typecheck
2. allow lenient on closure_js_test

I'm surprised that item 1 doesn't work. Let's first add some tests for that.

Item 2 seems like missing feature in closure_js_test but I don't understand why you need the js_binary part of the changes.

Regardless, let's separate those two issues into their own pull requests and discuss them separately.

michaelfig:
I may have been misusing Bazel, which made item 1 necessary.  I'm dropping it, and will open a new PR if I find that problem again.

The lenient (item 2) changes have been separated into #355.

PR #347, opened 2019-03-28
Upgrade closure compiler to latest release.

gkdn:
This also push internal changes to Java files since some of
these changes were done to make rules_closure compatible
with the Closure compiler.


davido:
@gkdn Just to let you know, this upgrade broke Gerrit Code Review Polymer build part. It seems that the problem is polymer-resin project. I reported this issue upstream: https://github.com/Polymer/polymer-resin/issues/7.

PR #349, opened 2019-03-29
Upgrade soy dependency

gkdn:
And remove no longer necessary shouldProvideRequireSoyNamespaces flag


gkdn:
@dslomov @laurentlb 
(wanted to make sure this is not out of your radar.)

PR #350, opened 2019-04-01
closure_js_template_library: add option to specify bidi_global_dir

### LINKED ISSUE #285: should_generate_soy_msg_defs requires BiDi directionality
Given the following `closure_js_template_library` rule : 

```
closure_js_template_library(
    name = "feedback_template",
    srcs = glob(
		[
			"**/*.soy"
		]
	),
    data = [],
    deps = [
        "//demo/epr/src/model"
    ],
    plugin_modules = [],
    should_provide_require_soy_namespaces = True,
    should_generate_soy_msg_defs = True,
    soy_msgs_are_external = False,
)
```

when I execute my build, I get the following exception : 

```
INTERNAL SOY ERROR.
Please open an issue at https://github.com/google/closure-templates/issues with this stack trace and repro steps
java.lang.IllegalStateException: If enabling shouldGenerateGoogMsgDefs, must also set bidi global directionality.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:504)
	at com.google.template.soy.jssrc.internal.JsSrcMain.genJsSrc(JsSrcMain.java:107)
	at com.google.template.soy.jssrc.internal.JsSrcMain.genJsFiles(JsSrcMain.java:151)
	at com.google.template.soy.SoyFileSet.compileToJsSrcFiles(SoyFileSet.java:1021)
	at com.google.template.soy.SoyToJsSrcCompiler.compile(SoyToJsSrcCompiler.java:223)
	at com.google.template.soy.AbstractSoyCompiler.compile(AbstractSoyCompiler.java:315)
	at com.google.template.soy.AbstractSoyCompiler.doMain(AbstractSoyCompiler.java:258)
	at com.google.template.soy.AbstractSoyCompiler.run(AbstractSoyCompiler.java:183)
	at com.google.template.soy.AbstractSoyCompiler.runMain(AbstractSoyCompiler.java:175)
	at com.google.template.soy.SoyToJsSrcCompiler.main(SoyToJsSrcCompiler.java:184)
Target //demo/epr/src/feedback:feedback failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 2.028s, Critical Path: 1.28s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

Proposed resolution would be to have the ability to provide the global bidi value expected by the soy compiler.

robfig:
This argument is passed directly through to SoyJsSrcCompiler.

Valid values are 1 (LTR) or -1 (RTL).

Fixes #285


gkdn:
LGTM.
It would be nice to have some tests (if we had one, this wouldn't have happened in the first place).

robfig:
Ping @laurentlb @dslomov for a LGTM

laurentlb:
Is there an easy way to add tests?

@gkdn, you can merge the change if you're happy with it

robfig:
I don't think I'm the right person to be creating a test suite for these rules.. I don't understand them like you do (or how to write test suites for bazel rules), and this is a small & straightforward fix that does not add significant risk or debt to the project.

gkdn:
The issue is if somebody send a patch to remove it won't break anything.
You don't need to write a complex unit test, just add a simple sample to tests that will exercise the code path and will break without your fix.

robfig:
I added a simple test, and it turned out to be easy! It's pretty basic, but is it what you had in mind?

closure/templates/closure_js_template_library.bzl
```diff
@@ -66,6 +68,7 @@ _closure_js_template_library = rule(
         "plugin_modules": attr.label_list(),
         "should_provide_require_soy_namespaces": attr.bool(default = True),
         "should_generate_soy_msg_defs": attr.bool(),
+        "bidi_global_dir": attr.int(),
```
gkdn: attr.int(values = [1, -1])

robfig: Cool, I didn't know about that. Updating

robfig: Oh, if I do that, then it makes the default of None==0 invalid. I could provide a default value of 1 which is probably what most callers would want, but up to you.

closure/templates/test/localized_test.js
```diff
@@ -0,0 +1,30 @@
...
+
+function testGreetLocalized() {
+  goog.soy.renderElement(goog.global.document.body,
+                         io.bazel.rules.closure.soy.greeter.greetLocalized,
+                         {name: 'Rob'});
+
+  var body = document.body;
+  goog.asserts.assert(body != null);
+  assertHTMLEquals('<p>Hello <b>Rob</b>!', body.innerHTML);
+}
```
gkdn: We ideally should test functionality of localization; ideally via replacing generated goog.getMsg calls somehow but this is an area of Closure that I'm not familiar at all so if you know how to do that it would be good.

Otherwise as an alternative, instead of having a closure test we can have a check that checks the output. Something like  following:

closure_js_template_library(
    name = "localized_soy",
    srcs = ["localized.soy"],
    should_generate_soy_msg_defs = True,
    bidi_global_dir = 1,
)

file_test(
    name = "localized_soy_generates_goog_get_msg",
    file = "localized.soy.js",
    regexp = "goog\.getMsg\(",
)

file_test(
    name = "non_localized_soy_not_generates_goog_get_msg",
    file = "greeter.soy.js",
    regexp = "goog\.getMsg\(",
    invert = True,
)

closure/templates/closure_js_template_library.bzl
gkdn: I guess making default one is good. As long as it doesn't break anything :)

closure/templates/test/localized_test.js
robfig: I'm actually not familiar with that either. My particular usage replaces the goog.getMsg calls with a custom translation lookup. I just updated it with your latter suggestion.

closure/templates/closure_js_template_library.bzl
robfig: Done. The tests pass at least

closure/templates/test/localized_test.js
gkdn: Do you really need to binary? Couldn't you just use the output of the closure_js_template_library target?

robfig: Oh interesting, I didn't realize that closure/templates/test:localized.soy.js was a target I could build to get a standalone library file! Will do

robfig: Updated

closure/templates/test/BUILD
```diff
@@ -34,6 +34,38 @@ closure_js_template_library(
     srcs = ["greeter.soy"],
 )
 
+
+closure_js_binary(
```
gkdn: pls remove and follow the same pattern

robfig: whoops, good catch. fixed

robfig:
Note that `bazel test closure/templates/test:*` fails for me at HEAD:

```
INFO: Analysed 100 targets (95 packages loaded, 915 targets configured).
INFO: Found 94 targets and 6 test targets...
ERROR: /Users/robfig/rules_closure/closure/templates/test/BUILD:27:1: Doing library-level typechecking of //closure/templates/test:person_proto failed (Exit 1) ClosureWorker failed: error executing command bazel-out/host/bin/java/io/bazel/rules/closure/ClosureWorker JsCompiler --checks_only --warning_level VERBOSE --jscomp_off reportUnknownTypes --language_in ECMASCRIPT_2017 --language_out ECMASCRIPT5 ... (remaining 48 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
"--lenient" is not a valid option
Sample usage: --compilation_level (-O) VAL --externs VAL --js VAL --js_output_file VAL --warning_level (-W) [QUIET | DEFAULT | VERBOSE]
```

gkdn:
Will take a look at the failure.

closure/templates/test/BUILD
```diff
@@ -34,6 +34,28 @@ closure_js_template_library(
...
+    file = ":localized.soy.js",
+    regexp = "goog.getMsg('Hello",
+)
+
+
+file_test(
+    name = "nonlocalized_test",
+    size = "small",
+    file = ":greeter.soy.js",
+    regexp = 'goog.getMsg("Hello',
```
gkdn: You quotes are inconsistent between two regexes. It doesn't seem right.

gkdn:
Sent https://github.com/bazelbuild/rules_closure/pull/357 for the failure.

closure/templates/test/BUILD
robfig: Good catch, updated. It actually is enough to check for absence of goog.getMsg, without caring about the argument.

PR #355, opened 2019-04-06
Lenient mode for closure_js_test

michaelfig:
This PR is necessary when building tests whose '*_test.js' and library sources both need `lenient = True`.

Without the `closure_js_binary` rule change, the build still fails with the errors that are suppressed in the new rule.

Thanks,
Michael.


michaelfig:
I found my `closure_js_binary` lenient failure was just because I missed some `lenient = True` settings from dependent libraries.  So, I removed that change, and added some tests.

closure/testing/test/BUILD
```diff
@@ -83,6 +83,22 @@ closure_js_test(
     ],
 )
 
+closure_js_library(
+    name = "lenient_lib",
+    srcs = ["lenient.js"],
+    lenient = True,
+)
```
gkdn: does the extra library serve any purpose?

closure/testing/test/lenient_test.js
```diff
@@ -0,0 +1,15 @@
...
+goog.scope(function() {
+    const Lenient = goog.require('rules_closure.Lenient');
+    
+    goog.testing.testSuite({
+      testMe: function() {
+        assertEquals(Lenient(), 123);
+      }
+    });
+    
+});  // goog.scope
```
gkdn: could you follow the simpler style:

testSuite({
  mytest() { ... }
});

Also could you document the part the you expect an error but will pass since it is lenient?

michaelfig: If i don't use a `goog.scope`, the compiler complains with:
```
closure/testing/test/lenient_test.js:5: ERROR - Closure primitive method goog.require may not be aliased  outside a module (ES module, CommonJS module, or goog.module)
const Lenient = goog.require('rules_closure.Lenient');
```

michaelfig: Okay, I've changed it to a goog.module.  FWIW, `entry_points` needed to be set, which isn't necessary with a goog.scope.

closure/testing/test/lenient_test.js
```diff
@@ -1,18 +1,14 @@
...
-});  // goog.scope
+const testSuite = goog.require('goog.testing.testSuite');
+const Lenient = goog.require('rules_closure.Lenient');
+
+testSuite({
+    testMe: function() {
+    // Without "lenient = True" on the lenient_test rule, this line
+    // fails with "ERROR - could not determine the type of this expression"
+    const val = Lenient();
+    assertEquals(val, 123);
```
gkdn: can you pls fix the formatting?

gkdn: And since you are there
```testMe: function() {..}``` --> ```testMe() { ... }```
And 
```on the lenient_test rule``` --> ```on the lenient_test target```

michaelfig: Okay, done!

Issue #359, opened 2019-04-15
ERROR: Bad --suppress value: JSC_DUPLICATE_NAMESPACE


pcj:
I'm trying to compile a generated javascript protobuf file, and getting this error that I'm trying to suppress:

```
bazel-out/k8-fastbuild/bin/external/com_github_stackb_grpc_js/js/pb/pb.i.js:2076: ERROR - Duplicate namespace: jspb.Message
goog.provide("jspb.Message");
^
  Codes: JSC_DUPLICATE_NAMESPACE
```

However, including this suppress code causes a different error `ERROR: Bad --suppress value: JSC_DUPLICATE_NAMESPACE`:

```
        suppress = [
            "JSC_DUPLICATE_NAMESPACE",
        ],
```

Trying to track this down but I suspect it may be related to https://github.com/bazelbuild/rules_closure/pull/347, but it's not so easy to bisect it.  

Any ideas?

gkdn:
Seems like a valid error if the js/pb/pb.js is providing a jspb.Message in addition to the one coming from the dep to ["@io_bazel_rules_closure//closure/protobuf:jspb"].

Looking at what com_github_stackb_grpc_js does, my rough guess is you are using protoc in a way that embeds jspb runtime into the generated code.

pcj:
Regardless of whether the error is valid or not should the compiler not respect the `suppress` value?  It seems like a bug somewhere if it reports the code but then does not recognize it as something that can be suppressed.

I think you are right about the duplicate namespace above.  I've removed `//js/pb` but now struggling with a different error:

```
bazel-out/k8-fastbuild/bin/external/com_github_stackb_grpc_js/js/grpc/stream/observer/call.i.js:5: ERROR - A file cannot be both a goog.provide'd file and a goog.module.
goog.module("grpc.stream.observer.UnaryCallObserver");
^
  Codes: JSC_MIXED_MODULE_TYPE
```

But again I cannot suppress `JSC_MIXED_MODULE_TYPE` for the generated `i.js` typecheck.

Yannic:
Seems like this is similar to #229

gkdn:
You can only suppress the ones and their aliases in 

https://github.com/bazelbuild/rules_closure/blob/50d3dc9e6d27a5577a0f95708466718825d579f4/java/com/google/javascript/jscomp/Diagnostics.java#L75

[...]

PR #360, opened 2019-04-17
Synchronize suppress values

pcj:
Copies valid suppression values from
commit 5e529e16e241a50f010ae868258f187c35f292e9

https://github.com/google/closure-compiler/blob/50713dceb16b369f8f317e5b49e8a9dec2b47496/src/com/google/javascript/jscomp/parsing/ParserConfig.properties

See #359


java/com/google/javascript/jscomp/Diagnostics.java
```diff
@@ -87,12 +89,15 @@
...
           "externsValidation",
+          "extraProvide",
           "extraRequire",
           "fileoverviewTags",
           "globalThis",
+          "googModuleExportNotAStatement",
           "invalidCasts",
-          "lateProvide",
           "legacyGoogScopeRequire",
+          "lateProvide",
```
gkdn: pls move up

java/com/google/javascript/jscomp/Diagnostics.java
```diff
@@ -87,12 +89,15 @@
...
+          "lintChecks",
```
gkdn: Looking at history, I think this was excluded on purpose; I don't want to dig why. Let's keep it for now.

java/com/google/javascript/jscomp/Diagnostics.java
```diff
@@ -87,12 +89,15 @@
           "duplicate",
           "es5Strict",
           "externsValidation",
+          "extraProvide",
```
gkdn: Looking at history, I think this was excluded on purpose; I don't want to dig why. Let's keep it for now.

java/com/google/javascript/jscomp/Diagnostics.java
```diff
@@ -77,6 +77,8 @@
           "accessControls",
           "ambiguousFunctionDecl",
           "checkDebuggerStatement",
+          "checkEventfulObjectDisposal",
```
gkdn: This is no longer needed.

java/com/google/javascript/jscomp/Diagnostics.java
pcj: Please clarify: keep it in the list or add another commit to remove this line?

java/com/google/javascript/jscomp/Diagnostics.java
pcj: Again I'm not sure if you're telling me to remove this line or not.

java/com/google/javascript/jscomp/Diagnostics.java
gkdn: Sorry for typo, I meant let's keep it that way. i.e. Don't include it in the list.

alexeagle:
@pcj seems like it's your turn on this one?

PR #361, opened 2019-04-17
closure_js_library: Add "library_level_checks" attribute

pcj:
This change adds a new boolean attribute to closure_js_library that allows the user to opt-out of "library level checks".

These checks --triggered by the named output %{name}_typecheck--, add a node in the action graph to execute JsChecker against an "i.js" file that represents an "ABI" of the library javascript.  While this is normally harmless, in certain circumstances the generated ijs file is syntactically incorrect and compilation errors against it represent a false-negative in the overall javascript compile action graph.

This attribute provides a way for users to opt-out of compiling the ijs file in these circumstances.
    
See #359


gkdn:
@pcj Could you provide an example where i.js is syntactically incorrect?

Also adding @blickly who added i.js support to JsChecker.
@blickly is this expected? what is the proper fix for this?

pcj:
OK, here's an example:

First, the error message(s):

```
bazel-out/k8-fastbuild/bin/external/com_github_stackb_grpc_js/js/grpc/stream/observer/call.i.js:5: ERROR - A file cannot be both a goog.provide'd file and a goog.module.
goog.module("grpc.stream.observer.UnaryCallObserver");
^
  Codes: JSC_MIXED_MODULE_TYPE

bazel-out/k8-fastbuild/bin/external/com_github_stackb_grpc_js/js/grpc/stream/observer/call.i.js:62: ERROR - A file cannot be both a goog.provide'd file and a goog.module.
goog.module("grpc.stream.observer.StreamingCallObserver");
^
  Codes: JSC_MIXED_MODULE_TYPE
```

I'm unsure where the filename `call.i.js` comes from as there is no equivalent source code file having that name.  By anyway, here's the source of that file (formatted):

```javascript
/** @fileoverview @typeSummary */
goog.provide("grpc.stream.observer.EventType");

/** @public @enum {string} */ grpc.stream.observer.EventType = {
  MESSAGE: 0,
  ERROR: 0,
  HEADERS: 0,
  TRAILERS: 0,
  COMPLETE: 0
};

/** @fileoverview @typeSummary */
goog.module("grpc.stream.observer.UnaryCallObserver");

const EventTarget = goog.require("goog.events.EventTarget");
const EventType = goog.require("grpc.stream.observer.EventType");
const GrpcStatus = goog.require("grpc.Status");
const Observer = goog.require("grpc.stream.Observer");

/**
 @extends {EventTarget}
 @implements {Observer<T>}
 @template T
 */
class UnaryCallObserver extends EventTarget {
  /**
   @param {!goog.promise.Resolver<T>} resolver
   */
  constructor(resolver) {}
  /** @protected */ handleResolve() {}
  /** @protected */ handleReject() {}
  /**
   @param {!Object<string,string>} headers
   @param {GrpcStatus} status
   @param {boolean=} opt_isTrailing
   @override
   */
  onProgress(headers, status, opt_isTrailing) {}
  /** @override */ onNext(value) {}
  /** @override */ onError(err) {}
  /** @override */ onCompleted() {}
  /**
   @protected
   @param {?Object<string,string>} headers
   @return {(GrpcStatus|undefined)}
   */
  getGrpcStatus(headers) {}
  /**
   @protected
   @param {?Object<string,string>=} headers
   @return {(string|undefined)}
   */
  getGrpcMessage(headers) {}
}

/** @private @type {T} */ UnaryCallObserver.prototype.value_;

/** @private @type {(!Object<string,string>|undefined)} */ UnaryCallObserver
  .prototype.trailers_;

/** @private @type {(!Object<string,string>|undefined)} */ UnaryCallObserver
  .prototype.headers_;

/** @private @type {(string|undefined)} */ UnaryCallObserver.prototype.message_;

/** @private @type {(GrpcStatus|undefined)} */ UnaryCallObserver.prototype
  .status_;

/** @protected @type {!goog.promise.Resolver<T>} */ UnaryCallObserver.prototype
  .resolver;

exports = UnaryCallObserver;

/** @fileoverview @typeSummary */
goog.module("grpc.stream.observer.StreamingCallObserver");

const UnaryCallObserver = goog.require(
  "grpc.stream.observer.UnaryCallObserver"
);

/**
 @extends {UnaryCallObserver<T>}
 @template T
 */
class StreamingCallObserver extends UnaryCallObserver {
  /**
   @param {!goog.promise.Resolver<?>} resolver
   @param {!function(T)} onMessage
   */
  constructor(resolver, onMessage) {}
  /** @override */ onNext(value) {}
  /** @override */ handleResolve() {}
}

/** @const @type {UnusableType} */ StreamingCallObserver.prototype.onMessage_;

exports = StreamingCallObserver;
```

The `JSC_MIXED_MODULE_TYPE` comes from https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/GatherModuleMetadata.java: 

```java
/**
 * Gathers metadata around modules that is useful for checking imports / requires and creates a
 * {@link ModuleMetadataMap}.
 */
public final class GatherModuleMetadata implements HotSwapCompilerPass {
  static final DiagnosticType MIXED_MODULE_TYPE =
      DiagnosticType.error("JSC_MIXED_MODULE_TYPE", "A file cannot be both {0} and {1}.");
```

This is a non-suppressible error code as it is not defined as a [DiagnosticGroup](https://github.com/google/closure-compiler/blob/edd9274aedc4d76759673b69d043189b68499aca/src/com/google/javascript/jscomp/DiagnosticGroups.java).

This error code is generated at this point where the module type is trying to be determined:

```java
    void moduleType(ModuleType type, NodeTraversal t, Node n) {
      checkNotNull(type);

      if (metadataBuilder.moduleType() == type) {
        return;
      }

      if (metadataBuilder.moduleType() == ModuleType.SCRIPT) {
        metadataBuilder.moduleType(type);
        return;
      }

      ambiguous = true;
      t.report(n, MIXED_MODULE_TYPE, metadataBuilder.moduleType().description, type.description);
    }
```

In this case, the `ModuleType type` argument didn't match the expected type of `metadataBuilder.moduleType(type)` and the module type was deemed "ambigious".

Looking at that ijs source code, I'm uncertain why that is.  Maybe it's the combination of something that looks like an es6 class and the "instance" fields that were represented like `StreamingCallObserver.prototype.onMessage_`, I'm not sure.

The problem is that this ambiguity only occurs within the JsChecker action on this ijs file.  Given the full source code, the compiler seems to do the right thing, the ambiguity is not present, and the compilation proceeds successfully.

Therefore, this option that I am proposing only serves to remove this JsChecker action for the given library rule only, allowing the user to get out of this situation.

gkdn:
As I mentioned earlier, the code you have pasted is a code that is bundled together. Each fileoverview is suppose to be a different file. That's not valid closure code and will cause lot different problems even the compiled version might work sometimes.

PR #344, opened 2019-03-26
Update protobuf version to 3.7.1

davido:
In: [1] the prerequisites were added to rules_closure for protobuf
upgrade to the version 3.7.0:

* Dependency on bazel_skylib
* Dependency on zlib

In this change the protobuf version is updated to 3.7.1.

[1] https://github.com/bazelbuild/rules_closure/pull/341

PR #363, opened 2019-04-25
Update repositories.bzl

laurentlb:
The repository now requires skylib and zlib, since the Protobuf update (#341, #344).


gkdn:
LGTM except the failing test.
Also it would be good to clarify "Due to recent changes" in the commit part. If recent changes are referred to rules_closure changes then it would be good to point to the relevant commit.

laurentlb:
Fixed the tests, updated the description.

PR #386, opened 2019-06-04
closure_js_template_library: add `defs` attr as a pass-through to SoyToJsSrcCompiler

### LINKED ISSUE #381: closure_js_template_library.should_provide_require_soy_namespaces does nothing
I'm using `closure_js_template_library` with an application that uses closure modules, so I need the generated code to use `goog.provide` / `goog.require`.  The documentation says that `should_provide_require_soy_namespaces` is passed through to the compiler, but in fact it does nothing because SoyToJsSrcCompiler had that flag removed at some point. 

That sounds ok, except I'm overriding it with a version of Soy that is a year or two old, and I can't update without fixing the various backwards incompatibilities that were introduced since then. The compiler I'm using still has that flag, and it defaults to false, so it's not possible for me to proceed without making some sort of change.

I wanted to ask if you would be open to a generic `args` parameter that acts as a passthrough to the compiler? This would solve this case as well as possible future cases when the flags change. There are also a good number of flags, so it's not infeasible to think that there are other valid use cases for wanting to get at them. 

Either way, I think it's worth updating the documentation on this parameter or possibly removing it entirely. I can do that if you give me a green light to make the pull request described above.

Thanks

https://github.com/bazelbuild/rules_closure#closure_js_template_library

robfig:
This also removes the vestigial `should_provide_require_soy_namespaces`

Fixes #381

### DIFF
--- a/closure/templates/closure_js_template_library.bzl
+++ b/closure/templates/closure_js_template_library.bzl
@@ -32,6 +32,10 @@ def _impl(ctx):
         args += ["--bidiGlobalDir=%s" % ctx.attr.bidi_global_dir]
     if ctx.attr.plugin_modules:
         args += ["--pluginModules=%s" % ",".join(ctx.attr.plugin_modules)]
+    for arg in ctx.attr.defs:
+        if not arg.startswith("--") or (" " in arg and "=" not in arg):
+            fail("Please use --flag=value syntax for defs")
+        args += [arg]
     inputs = []
     for f in ctx.files.srcs:
         args.append("--srcs=" + f.path)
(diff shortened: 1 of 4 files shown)


closure/templates/closure_js_template_library.bzl
```diff
@@ -32,6 +32,8 @@ def _impl(ctx):
         args += ["--bidiGlobalDir=%s" % ctx.attr.bidi_global_dir]
     if ctx.attr.plugin_modules:
         args += ["--pluginModules=%s" % ",".join(ctx.attr.plugin_modules)]
+    for arg in ctx.attr.defs:
```
gkdn: could you add validation similar to jscompiler one?

robfig: good catch, added!

robfig:
Thanks! FYI, I do not have permission to merge it

PR #391, opened 2019-06-10
closure_js_library: fix: 'exports' now works in a data attr

robfig:
This adds a test that reproduces an issue I'm running into. I'm not sure what the fix is, so I figured I would raise it and see if you knew if this was expected or could provide any guidance on what's missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -570,7 +570,7 @@ The documentation on using Closure Templates can be found
 For additional help on using some of these attributes, please see the output of
 the following:
 
-    bazel run @io_bazel_rules_closure//third_party/java/soy:SoyToJsSrcCompiler -- --help
+    bazel run @com_google_template_soy//:SoyToJsSrcCompiler -- --help
 
 #### Implicit Output Targets
 
@@ -867,7 +867,7 @@ This rule can be referenced as though it were the following:
 
 - **defs:** (List of strings; optional) Specifies additional flags to be passed
   to the Closure Stylesheets compiler. To see what flags are available, run:
-  `bazel run @io_bazel_rules_closure//third_party/java/csscomp:ClosureCommandLineCompiler`
+  `bazel run @com_google_closure_stylesheets//:ClosureCommandLineCompiler -- --help`
 
 
 ## closure\_js\_proto\_library
--- a/closure/compiler/closure_js_library.bzl
+++ b/closure/compiler/closure_js_library.bzl
@@ -432,6 +432,7 @@ def _closure_js_library(ctx):
                 transitive = [
                     collect_runfiles(unfurl(ctx.attr.deps, provider = "closure_js_library")),
                     collect_runfiles(ctx.attr.data),
+                    collect_runfiles(library.exports),
                 ],
             ),
         ),
--- a/closure/compiler/test/BUILD
+++ b/closure/compiler/test/BUILD
@@ -214,3 +214,26 @@ file_test(
     file = "extern_bad_bin-stderr.txt",
     regexp = "actual parameter 1 of omg.im_an_extern does not match formal parameter",
 )
+
+# TEST: closure_js_library targets from exports should be provided when used in
+# a data attr.
+
+closure_js_library(
+    name = "external",
+    srcs = ["external.js"],
+    data = ["external2.js"],
+)
+
+closure_js_library(
+    name = "exports_data",
+    exports = [":external"],
+)
+
+sh_test(
+    name = "exports_data_test",
+    srcs = ["exports_data_test.sh"],
+    data = [
+        # including external.js / external2.js through a rollup target via exports should work.
+        ":exports_data",
+    ],
+)
--- a/closure/compiler/test/exports_data_test.sh
+++ b/closure/compiler/test/exports_data_test.sh
@@ -1,0 +1,2 @@
+runfiles=closure/compiler/test
+[[ -e "$runfiles/external.js" && -e "$runfiles/external2.js" ]]
--- a/closure/repositories.bzl
+++ b/closure/repositories.bzl
@@ -202,9 +202,8 @@ def args4j():
 def bazel_skylib():
     http_archive(
         name = "bazel_skylib",
-        sha256 = "bbccf674aa441c266df9894182d80de104cabd19be98be002f6d478aaa31574d",
-        strip_prefix = "bazel-skylib-2169ae1c374aab4a09aa90e65efe1a3aad4e279b",
-        urls = ["https://github.com/bazelbuild/bazel-skylib/archive/2169ae1c374aab4a09aa90e65efe1a3aad4e279b.tar.gz"],
+        sha256 = "2ef429f5d7ce7111263289644d233707dba35e39696377ebab8b0bc701f7818e",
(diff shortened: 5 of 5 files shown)


closure/testing/test/BUILD
```diff
@@ -111,6 +111,46 @@ closure_js_test(
...
+closure_js_test(
+    name = "external_code_test",
+    srcs = ["external_code_test.js"],
+    html = "external_code_test.html",
+    entry_points = [
+        "rules_closure.ExternalCodeTest",
+    ],
+    data = [
+        # including external.js through a rollup target via exports should work.
+        ":all_external",
```
Yannic: Does it work if you add `:external`?

I'm not very familiar with how the data attribute works. Does it consider `files`, `runfiles`, or both as input? If `runfiles` are used, then it should be easy to add. If it uses `files`, then it also shouldn't work if `:external` is a direct dependency, and this is working as intended.

https://github.com/bazelbuild/rules_closure/blob/308b05b2419edb5c8ee0471b67a40403df940149/closure/compiler/closure_js_library.bzl#L428

robfig: Yes, it works if I reference the library directly. I'm afraid I don't know how it works either.

Yannic: Okay, then, AFAICT, this is indeed a bug.
You should be able to fix it by adding `library.exports` to [this list](https://github.com/bazelbuild/rules_closure/blob/308b05b2419edb5c8ee0471b67a40403df940149/closure/compiler/closure_js_library.bzl#L432) :)

robfig: Nailed it! That fixed the test. I've updated this pull request.

robfig:
@dslomov @gkdn @laurentlb Please review, this is a small bug fix with unit test.

robfig:
Good point, externs are not needed. Please take another look

closure/testing/test/BUILD
```diff
@@ -111,6 +111,40 @@ closure_js_test(
     ],
 )
 
+# TEST: closure_js_library targets from exports should be provided when used in
+# a data attr.
+
+closure_js_library(
+    name = "external",
+    srcs = ["external.js"],
```
gkdn: I think this part is little bit misleading. You are not trying to compile the code. That's why I recommended putting it on the data.

closure/testing/test/BUILD
```diff
@@ -111,6 +111,40 @@ closure_js_test(
...
+)
+
+closure_js_library(
+    name = "all_external",
```
gkdn: Let's put a more describing name; point is exporting right?

closure/testing/test/BUILD
```diff
@@ -111,6 +111,40 @@ closure_js_test(
...
+    entry_points = [
+        "rules_closure.ExternalCodeTest",
+    ],
+    data = [
+        # including external.js through a rollup target via exports should work.
+        ":all_external",
+    ],
+    # NOTE: I'm not sure why this is necessary.
+    suppress = [
+        "reportUnknownTypes",
```
gkdn: Is this still needed? What is unknown here?

closure/testing/test/BUILD
robfig: In my use case, I have a closure_js_library for each of my third party libraries, and they are compiled together using forgiving settings to create a vendor bundle. However, code that requires those libraries as externs needs to include them as data for their tests. I can change this test if you want, but then it wouldn't test my actual usage of this feature.

closure/testing/test/BUILD
robfig: `exported_external`?

closure/testing/test/BUILD
robfig: It's not still needed. Removing

closure/testing/test/BUILD
gkdn: > In my use case, I have a closure_js_library for each of my third party libraries, and they are compiled together using forgiving settings to create a vendor bundle. However, code that requires those libraries as externs needs to include them as data for their tests.

In that care you probably want a closure_js_binary and depend that as data in the test.

> I can change this test if you want, but then it wouldn't test my actual usage of this feature.

Your use case sounds non-trivial but the test should reflect basic underlying functionality/bug for maintenance purposes.

closure/testing/test/BUILD
gkdn: exports_data?

closure/testing/test/BUILD
gkdn: > In my use case, I have a closure_js_library for each of my third party libraries, and they are compiled together using forgiving settings to create a vendor bundle. However, code that requires those libraries as externs needs to include them as data for their tests.

In that care you probably want a closure_js_binary and depend that as data in the test.

> I can change this test if you want, but then it wouldn't test my actual usage of this feature.

Your use case sounds non-trivial but the test should reflect basic underlying functionality/bug for maintenance purposes.

gkdn: > In my use case, I have a closure_js_library for each of my third party libraries, and they are compiled together using forgiving settings to create a vendor bundle. However, code that requires those libraries as externs needs to include them as data for their tests.

In that care you probably want a closure_js_binary and depend that as data in the test.

> I can change this test if you want, but then it wouldn't test my actual usage of this feature.

Your use case sounds non-trivial but the test should reflect basic underlying functionality/bug for maintenance purposes.

closure/testing/test/BUILD
gkdn: exports_data?

robfig: updated

gkdn:
Github is acting on me. Sorry for duplicates.

I noticed that the test is in in the wrong location. Need to be moved under closure/compiler/test/<scenario> and checking if something is on runfiles is easier to be done via sh_test:
[[ -e third_party/bazel_rules/rules_closure/closure/compiler/test/<scenario>/<data_file_name> ]]

closure/testing/test/BUILD
robfig: Yeah, those are still built with plovr, I suppose eventually we would use the binaries instead. I'll update it to just use data.

robfig: Having a data-only closure_js_library target is not valid:

```
closure_js_library(
    name = "external",
    data = ["external.js"],
)
```

results in
```
Either 'srcs' or 'exports' must be specified
```

robfig:
Sure thing. What do you want me to do with respect to using the `data` attribute instead of `srcs` for this test? Should I be using a different target type? 
https://github.com/bazelbuild/rules_closure/pull/391#discussion_r294370175

gkdn:
> Sure thing. What do you want me to do with respect to using the `data` attribute instead of `srcs` for this test? Should I be using a different target type?
> [#391 (comment)](https://github.com/bazelbuild/rules_closure/pull/391#discussion_r294370175)

Let's make it general runfiles test and include but data and js srcs and verify both exists in the sh_test.

robfig:
Updated, PTAL

gkdn:
Thanks. Looks good.

robfig:
Updated

PR #392, opened 2019-06-13
Update bazel_skylib version to 0.8.0

meteorcloudy:


meteorcloudy:
TensorFlow fails after I upgrade rules_closure with the following error:
```
ERROR: Analysis of target '//tensorflow/tools/pip_package:build_pip_package' failed; build aborted: invalid registered toolchain '@io_bazel_rules_docker//toolchains/docker:default_windows_toolchain': error loading package '@io_bazel_rules_docker//toolchains/docker': Unable to load file '@bazel_skylib//:bzl_library.bzl': file doesn't exist
```

Turned out it's because rules_closure is using an old version of bazel_skylib, which is not compatible with other TF dependencies that also depends on bazel_skylib.

PR #402, opened 2019-07-09
Migrate org.mockito.Matchers#any* to org.mockito.ArgumentMatchers

TimvdLippe:
The former is deprecated and replaced by the latter in Mockito 2. However, there is a
functional difference: ArgumentMatchers will reject `null` and check the type
if the matcher specified a type (e.g. `any(Class)` or `anyInt()`). `any()` will
remain to accept anything.

This reflects the internal changes 257191326 and 257195762


dslomov:
Hey looks like we need to update the mockito in WORKSPACE file as well

TimvdLippe:
Opened #403

PR #405, opened 2019-07-18
Update Closure Library and Closure Templates

ribrdb:
Because the releases of the library and templates aren't in sync, the latest release of the Closure Library isn't compatible with any released version of Closure Templates. So I used the latest release of Closure Templates and a recent master commit of Closure Library.


sgammon:
Eh.... but you do know template dep handling changed, right? It’s using a new SoyHeaderCompiler now

I’ve got a PR I can contribute soon to make it work against closure-templates at HEAD.

sgammon:
Just making sure you caught that because it snagged us good

ribrdb:
No I didn't see that. Just updating the library seems to be working for us, although I don't think we have any templates that have dependencies.

davido:
@ribrdb I was trying to upgrade protobuf to recent release:  3.10.0-rc1 in this PR: [1] and realized that it depends on recent closure library (v20190729). Can you upgrade this PR?

[1] https://github.com/bazelbuild/rules_closure/pull/413

closure/repositories.bzl
```diff
@@ -728,10 +728,10 @@ def com_google_template_soy():
         name = "com_google_template_soy",
```
Yannic: @sgammon is correct, `closure-templates` changed how deps are passed to the compiler, so this will break everyone using templates with dependencies. Given that it's a larger change to adapt `rules_closure`, can we revert this and only upgrade closure-library in this PR?

(On a slightly related note: Why don't we have tests for this?)

closure/repositories.bzl
```diff
@@ -680,11 +680,11 @@ def com_google_javascript_closure_library():
...
         urls = [
-            "https://mirror.bazel.build/github.com/google/closure-library/archive/v20190415.tar.gz",
-            "https://github.com/google/closure-library/archive/v20190415.tar.gz",
+            "https://mirror.bazel.build/github.com/google/closure-library/archive/82b285261eedce6a52bf1d3b1667e16c6d10fe1e.tar.gz",
+            "https://github.com/google/closure-library/archive/82b285261eedce6a52bf1d3b1667e16c6d10fe1e.tar.gz",
         ],
-        sha256 = "b92f9c2b81adeb06b16e0a5d748baf115eeb58884fe864bec3fe8e23a7cf7d14",
-        strip_prefix = "closure-library-20190415",
+        sha256 = "61cbeac7c403b8d591f2edaa717faa90048787012de55442dad3c9d1f2dab7b2",
+        strip_prefix = "closure-library-82b285261eedce6a52bf1d3b1667e16c6d10fe1e",
```
Yannic: I'm with @davido, can you update this PR for `v20190819`.

Someone will also need to upload the archive to `mirror.bazel.build`.

ribrdb:
I don't think that will work. I'm pretty sure I only updated closure templates because things didn't compile with the new closure library and closure templates 2019-03-11. I'm out of the office so I might not be able to update this for a few days.

closure/repositories.bzl
Yannic: Actually, I'm not sure why, but this doesn't seem to break `closure_{js,java}_template_library`. Apparently, we never use `--deps`.

ribrdb:
I've tested this with #414 and all the tests still pass.
I can't update to v20190819 without also updating protocol buffers. Should I rebase on top of #413 ? Is that one ready?

davido:
> Is that one ready?

I mean yes, but there is one additional request to explicitly add repository for rules_java, even though it works in 0.29.x and upcoming 1.0.x releases without it.

davido:
Wait a moment I will explicitly add the `rules_{cc,java}` repositories as requested in the review.

Yannic:
Can you try `v20190729`? It has `goog.crypt.base64.Alphabet` that is needed for #413 (https://github.com/google/closure-library/commit/74991c9a6390c8d1e2d80cbfe4b6355d8a522c2c#diff-2b2337e9b17a1a7a89e05a607ea6cc78), but doesn't remove the old API yet (https://github.com/google/closure-library/commit/0b71c13c3cc29872893f92a9a902def33b7b9de8#diff-2b2337e9b17a1a7a89e05a607ea6cc78).

davido:
@ribrdb @Yannic I added `rules_{cc,java}` in #413. It should be ready now.

ribrdb:
Ok, this is with v20190729. I'm not sure if that unblocks #413.

Issue #411, opened 2019-09-04
rules_closure violates Closure Compiler visibility for its Java packages


concavelenz:
rules_closure/java/com/google/javascript/jscomp/ places files within the Closure Compiler package allowing it to violate package visibility and use internal APIs.

There are no guarantees about internal APIs and makes it difficult for the Closure Compiler authors to avoid breaking these rules.

PR #416, opened 2019-09-12
Replace uses of  public fields with their associated getter methods.

nreid260:


Yannic:
Thanks! I'm assuming this is working towards #411.
What do we need to get the tests to pass? Update closure-compiler to `v20190909`?

nreid260:
It's more like an example of the difficulty described by that issue. The fact that I need to make this PR at all is what we consider problematic.

I think updating to `v20190909` is all that should be needed. That version has both the getters and the public fields.

Yannic:
We updated closure compiler to `v20190909`. Can you rebase so we can see if the tests pass?

nreid260:
Done

PR #422, opened 2019-09-14
Retire Travis in favor of Buildkite

Yannic:
This PR retires Travis in favor of Buildkite.


Yannic:
Seems like Bazelisk is running into GitHub rate limiting on Travis.

//cc @philwo

philwo:
@Yannic I can think of three ways to fix this:
- Don't use Travis ;)
- Create a non-privileged GitHub user, get an API token for it, pass that token to Bazelisk via the `BAZELISK_GITHUB_TOKEN` environment variable. Then you should no longer run into rate limiting.
- Implement https://github.com/bazelbuild/bazelisk/issues/87, then we no longer use the GitHub API to figure out the latest release, so rate limiting will no longer happen.

Yannic:
I'm voting for removing Travis. It's slow and doesn't provide anything we can't get from Buildkite.

@gkdn @dslomov @laurentlb WDYT?

alexeagle:
Sounds great @Yannic 

Can you check that all the tests run on Travis are also run on buildkite, so we know we don't lose any coverage? And update the PR to just remove the travis config.

Yannic:
We're running the same tests on Buildkite as we run on Travis (`//...`), and, IIUC, sandboxing is the default these days.

Someone with write (or even admin) access needs to remove that Travis is a required check for PRs before we can delete the config. @dslomov @laurentlb Can one of you handle this?

philwo:
@Yannic Done :)

Yannic:
@alexeagle Done. ptal

PR #424, opened 2019-09-14
WIP: Create soy_library

Yannic:
Blocked on: https://github.com/google/closure-templates/issues/193

//cc @sgammon

### DIFF
--- a/closure/private/closure_common.bzl
+++ b/closure/private/closure_common.bzl
@@ -1,0 +1,19 @@
+# Copyright 2019 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//closure/private/utils:soy.bzl", "soy")
+
+closure_common = struct(
+    soy = soy,
+)
--- a/closure/private/providers.bzl
+++ b/closure/private/providers.bzl
@@ -1,0 +1,30 @@
+# Copyright 2019 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+SoyInfo = provider(
+    doc = "Encapsulates information provided by soy_library.",
+    fields = {
+        "direct_headers": "A set of soy header files of the direct sources. " +
+                          "If the library is a proxy library that has no " +
+                          "sources, it contains the direct_headers from this " +
+                          "library's direct deps.",
+        "direct_sources": "A set of soy sources from the 'srcs' attribute.",
+        "proto_descriptor_sets": "A set of FileDescriptorSet files of all " +
+                                 "dependent proto_library rules.",
+        "indirect_headers": "A set of soy header files from all dependent " +
+                            "soy_library rules, but excluding direct headers.",
+        "indirect_sources": "A set of soy sources of all dependent" +
+                            "soy_library rules, but excluding direct sources.",
+    },
+)
--- a/closure/private/rules/BUILD
+++ b/closure/private/rules/BUILD
@@ -1,0 +1,13 @@
+# Copyright 2019 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/closure/private/rules/soy_library.bzl
+++ b/closure/private/rules/soy_library.bzl
@@ -1,0 +1,66 @@
+# Copyright 2019 The Closure Rules Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//closure/private:closure_common.bzl", "closure_common")
+load("//closure/private:providers.bzl", "SoyInfo")
+
+def _soy_library_impl(ctx):
+    soyh = ctx.actions.declare_file("{}.soyh.bin".format(ctx.label.name))
+    return [
+        DefaultInfo(
+            files = depset([soyh]),
+        ),
+        closure_common.soy.compile(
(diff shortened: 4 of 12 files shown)


sgammon:
this looks fantastic @yannic and much cleaner than my approach

sgammon:
can this get merged? :)

sgammon:
if so i can begin contributing my fork stuff downstream, which is built on top of this and other PRs.

gkdn:
Thanks.

sgammon:
I'm using this change downstream in a fork @gkdn. I'd love to get it updated so we can eventually track with mainline.

sgammon:
Would you guys welcome an updated contribution here? @Yannic may be busy in which case I'm happy to roll up my sleeves and give it a try.

gkdn:
I think this is related to one of wrong term direction discussions for rules_closure.

I think this repo should solely focus on providing a slim compatibility layer around closurized JS in Bazel which is the JavaScript library and JS provider aspect.

With that I think the right long term approach here is to actually taking soy / closure template related pieces to its own repository and maintain that independently (we actually filter whole closure template pieces while syncing repo back to Google).

However I am happy to hear any counter arguments to this.

sgammon:
@gkdn i can see your point and i understand, that, from J2CL et. al's perspective, Soy and GSS might be seen as "extras."

`rules_closure` is, in essence, the _only_ working full Closure stack outside of Google (or closest to it, minus PRs such as this one), aside from building with CLI invocations and a Makefile. if it were simply up to date with some of these PRs merged, it would be a fantastic stack to use if you're building CSR single-js-file apps from Bazel.

it's funny that i should be asked to answer this question. allow me to explain why.

i first adopted GCP when it was just App Engine in 2008. i sat in the front row of the first Google I/O. i had come in that morning by train. i think i skipped high school classes to attend. this is where Google first entered my life: at the earliest point in my development career.

for most of the time since, i have been a 100% google dev. i built all of my early software on the GAE Python runtime, including suffering the transition from 2.5 -> 2.7 for multithreading, and so on. i tried my very best to build "like Google does," under the theory that, while different from everything else, following Google's path would mean Google performance, and with the same built-in process innovation that Googlers enjoy. obviously this necessitated a rejection of the status quo, but Google does that every day, and as long as Google's platform "is the open web," then our interests are aligned. right?

building "like google does" evolved from Guido's NDB, to [ProtoRPC](https://github.com/sgammon/AppEngine-Toolkit/issues/2) (!), to Protobuf, to Java, to Closure, to Bazel. years later, in finding Bazel, the miracle of google's stack finally became fully clear: with Bazel managing the dependencies for protos and Soy and JS sources, all works harmoniously as i had dreamed back in 2008-2010. i believed in `rules_closure` so much, i wrote a [framework on top of it](https://github.com/sgammon/elide/tree/master).

so, I want to humbly ask that you consider keeping these tools together, because that last working Closure stack is meaningful to the broader internet, and Google. certainly more than might seem from inside, where these tools are facing stronger and stronger deprecation signals over time. Google Docs and MS365 are (were?) _both_ built by Closure Compiler, so it is not going anywhere, but i struggle to think of reasons that will prevent Soy or GSS from being lost to time.

all of that being said, building "like google does" is actually something i am doing less of, not more. you may remember the bugs i filed that helped create J2CL-proto, or this PR, and so on, and those were, perhaps, the last moments of that era in my career. why? because of decisions exactly the one you are proposing here -- [loads](https://github.com/material-components/material-components-web/issues/4839) [of](https://github.com/bazelbuild/rules_nodejs/issues/2403) [them](https://github.com/lit/lit-element/issues/1109), [perhaps](https://github.com/angular/tsickle/issues/1040) [hundreds](https://github.com/bazelbuild/rules_nodejs/issues/2767) [over](https://github.com/google/elemental2/pull/120) [time](https://github.com/firebase/firebase-js-sdk/issues/1861), which have taught me consistently that google actually has no interest at all in supporting developers who choose to develop "like google does" (except, of course, the ones who work at Google).

i realize it is unfair to hold _you_ to account for these hundreds of decisions (despite participating in a few of them). i also realize it is unfair to hold [_any_ particular Googler to account](https://carlmastrangelo.com/blog/why-does-grpc-insist-on-trailers) in influencing an entity as large as Google, and i am fully sure that Googlers fight The Good Fight every day to support open source. i can't even fault your motivations, because, ultimately, you have stakeholders to answer to like all of us do, and they don't care about Closure's accoutrements, so why keep them? i get it. you would be abdicating your essential duty as an engineer to pretend otherwise.

**i would love for these tools to stay. it would be useful for these tools to stay.** but, i know better now than to stake a library or framework on it, and i have a fork that works, having dispensed with any patience long ago. most of the code i write is now targeted at Kotlin/JS, which works really well with React. @jart and @Yannic aren't active anymore, and @jart is writing [amazing shit](https://justine.lol/make/) outside of Google that fights this innovation dilemma at an ideological level. the `rules_closure` era has passed, so, if _you_ think they should stay and are willing to update them _every now and then_, which is frankly a _minor_ cost to Google that should justify itself, then yes, they should stay, so that users who want to remain using closure, or try it out for posterity, still can.

resources are always finite, and with such a minor user base outside of Google, i understand this _appears_ costly. but i'm hopeful nonetheless that, someday, these tools can get a shot at a second life. with Kotlin's new IR compiler, perhaps a bridge could be written to Closure, which makes the strong security guarantees of Soy more attractive to a new and broader audience. it is reasonable to argue that the full Closure stack could be characterized as the _most road-tested framework of all time_, and that, to me, alone, justifies keeping it.

in other words, the internet is trending toward _stronger_ typing in the browser, rather than weaker typing, and the Closure stack may appeal in new and different ways if we are wise enough to keep it around for open source use.

of course, it is tempting to argue that newer technologies from Google should _support Google's [own tech](https://github.com/google/elemental2/issues/119) first class_, [especially _outside_ of Google](https://github.com/material-components/material-web/issues/2453). but seeing as that is not likely to be the outcome anyway (regardless of what me, you, or others do), i would not presume to offer it as my answer.

sgammon:
@gkdn, sorry, one last point. i _do_ think this stack of tools is deeply useful -- Soy and GSS included -- for developing chrome extensions, and should be kept on that basis as well.

even if the closure stack isn't as competitive as others at targeting regular web runtimes, the Chrome externs make extensions really safe and easy to write from Bazel, which should perhaps be a consideration here before dispensing with them entirely. i'm not actually sure _how_ i'd achieve parity in that context _without_ Closure (to include the full set, most of all Soy).

gkdn:
(And FWIW, since you linked above to https://github.com/google/elemental2/issues/119; we proposed a good alternative and you said you will follow up with such a PR but didn't 🙂)

sgammon:
@gkdn fair point re/moving-not-deprecating and i am not against that at all in principal, but to clarify, i'm concerned about what that move would mean in practice.

to an outside user, you're right, the story is very fragmented, as evidenced by the long journey it took me (as a motivated sample case) to find out how the whole stack is _supposed_ to fit together.

my concern is that, by further separating these tools, they would become even more fragmented w.r.t. their external story, and would see lagging dependencies and feature support because they would no longer be part of a project that enjoys the support `rules_closure` does due to downstream dependencies (i.e. J2CL, elemental2, and so on).

there is, of course, the practical consideration of Soy et all. being coupled tightly to Closure Library and the Compiler, but moving Soy Library out of this repo is frankly a solid anti-case to my point. that was a beneficial change and was probably more complex than moving Soy and GSS.

if you are confident that moving the tools wouldn't result in lagging dependencies and feature support, then I'm all for it. i think you are in a better position to evaluate this than me, considering you have visibility into Google's use case and forward-looking posture, which seems to be the prevailing wind in most circumstances.

sgammon:
@gkdn, yes, this may shock you, but we have even less bandwidth to add features to google's software _outside_ google than google has inside google

unfortunately, as crazy as it may sound, the presence (or lack thereof) of Chrome externs, as described in that ticket, ended up governing whether Bazel was even an option for us for that product. development has stalled as a result. some ideas never make it to market -- for all sorts of reasons, but, in Bazel and Closure's case, more often than not because the tech could not be made to work together (again, outside of Google).

so yes. i may owe you guys a PR, but my use case that drove it was negated by the idea that we, a small startup, would be supporting Chrome's integration with Elemental2, which is a little crazy.

could it be automated? maybe. but spending the time to figure that out is not something i am equipped to defend to investors.

sgammon:
in other words, in my humble opinion, when a user comes along to google and says, "please, could we get this tiny feature? there is no use case inside google yet, we know, but if it exists, we will build an entire business on it and adopt aggressively and contribute as long as we are around"

the answer should be, "of course. the stated and actual intent is for you to build a business on top of google, so we will expend an indirect and comparatively light cost for a major gain: a lifelong customer on GCP who is tightly coupled to our technologies. it will perform better in GCP, having been designed there, so you will stay paying us. in exchange for the business risk of that coupling, we will help keep your software functional"

but this has been violated so many times. from Chrome refusing to support gRPC, to Polymer's ultimate deprecation, and on and on. merely realizing that we would be fighting headwinds was enough to see the writing on the wall as, "don't build a business here; if it isn't in Google's immediate and obvious interest, no force on earth can make it happen."

again, not at all your fault, just trying to explain why moving it _looks_ like Google Reader instead of, say, Polymer to Lit.

sgammon:
fwiw that Chrome idea has seen a revival recently, so maybe i will try to go land that PR.

PR #432, opened 2019-10-07
Protobuf: Remove java 7 compatibility mode

davido:
This is a workaround to suppress this annoying warning:

  warning: -parameters is not supported for target value 1.7. Use 1.8 or later.

wenn building protobuf_java. The full story can be seen in this Bazel
issue: [1]. I tried hard to avoid the need for patching Protobuf on
fetch, by trying to remove Java 7 compatibility mode upstream, that was
rejected: [2]. Moreover, I tried hard and spent quite some time to fix
the problem in Bazel: [3], in non-invasive way, but this PR was as well
rejected. The next natural approach is to just patch the Protobuf source
during the fetch operation.

Needless to say, that the patching during the fetch has its own
disavdantages, that the patch would probably need to be updated on
every Protobuf upgrade. But what can you do? The developers are
complaining and filling the issues in every issue tracker, as here in
Gerrit Code Review project: [4].

[1] https://github.com/bazelbuild/bazel/issues/8772
[3] https://github.com/bazelbuild/bazel/pull/9494
[3] https://github.com/protocolbuffers/protobuf/pull/6711
[4] https://bugs.chromium.org/p/gerrit/issues/detail?id=11102

### DIFF
--- a/closure/protobuf_drop_java_7_compatibility.patch
+++ b/closure/protobuf_drop_java_7_compatibility.patch
@@ -1,0 +1,39 @@
+diff --git a/BUILD b/BUILD.bazel
+index efc3d8e..2f27833 100644
+--- a/BUILD
++++ b/BUILD.bazel
+@@ -653,10 +653,7 @@ java_library(
+     ],
+     javacopts = select({
+         "//:jdk9": ["--add-modules=jdk.unsupported"],
+-        "//conditions:default": [
+-            "-source 7",
+-            "-target 7",
+-        ],
++        "//conditions:default": [],
+     }),
+     visibility = ["//visibility:public"],
+ )
+@@ -754,10 +751,7 @@ java_library(
+     ],
+     javacopts = select({
+         "//:jdk9": ["--add-modules=jdk.unsupported"],
+-        "//conditions:default": [
+-            "-source 7",
+-            "-target 7",
+-        ],
++        "//conditions:default": [],
+     }),
+     visibility = ["//visibility:public"],
+ )
+@@ -767,10 +761,6 @@ java_library(
+     srcs = glob([
+         "java/util/src/main/java/com/google/protobuf/util/*.java",
+     ]),
+-    javacopts = [
+-        "-source 7",
+-        "-target 7",
+-    ],
+     visibility = ["//visibility:public"],
+     deps = [
+         "protobuf_java",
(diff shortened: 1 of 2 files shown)


alexeagle:
Thanks for the detailed explanation. I agree that you've exhausted the better avenues for fixing this so I'm fine carrying a patch.

Let's discuss separately whether I can migrate gerrit off of rules_closure, we can probably build the same thing just using the google-closure-compiler npm package.

davido:
@alexeagle Thanks for the approval!

> Let's discuss separately whether I can migrate gerrit off of rules_closure, we can probably build the same thing just using the google-closure-compiler npm package.

+1. Gerrit Front End team located in Google@MUC is working to upgrade
Gerrit Code Review FE from Polymer 1 to Polymer 2 and it is almost ready,
and the next Gerrit release Gerrit 3.1 would be switched to Polymer 2.

Moreover we would like to re-deisgn current Gerrit frontend build tool chain machinery.
This is the gerrit issue: [1]. Your help would be greatly appreciated! 

[1] https://bugs.chromium.org/p/gerrit/issues/detail?id=11226

alexeagle:
@davido sure, maybe we should chat about it in slack.bazel.build - I have cloned the gerrit repo and started looking at how to migrate but it's very complex. I'm starting with https://github.com/jleyba/js-dossier as a simpler example usage to find a simpler and better supported way.

davido:
@alexeagle I will attend upcoming Gerrit Hackathon in Sunnyvale (11th November - 15th November). May be you (or someone else from your team) could join us for a meeting to discuss the possibilities?

I don't know who will attend from FE team from Google@MUC, though.

alexeagle:
That sounds perfect. I remember meeting Shawn P long ago at a Jenkins hackathon around here when I think the idea for Gerrit was just starting. Let me know when is a good time during that hackathon to come by?

davido:
@lucamilanesio Can you help to plan 2-3 hours pn site meeting with @alexeagle during the hackathon? 

And send the invites to Ole Rehmsen and Ben Rohlfs, both from FE team from Google@MUC?

lucamilanesio:
Sure, @alexeagle what date / time would be best for you? The hackathon will take place in the GerritForge office in Sunnyvale, `100 S Murphy Ave Suite 200, Sunnyvale, CA 94086, United States`

PR #437, opened 2019-11-01
Upgrade Closure compiler to version v20191027.

rluble:


rluble:
It seems that there is more to updating the version of closure compiler than I had anticipated.

Issue #419, opened 2019-09-13
Files missing mirror.bazel.build


Yannic:
See e.g. https://buildkite.com/bazel/rules-closure-closure-compiler/builds/558#6c9b9967-a67c-4151-9283-06179d6e47bd/239-250

```
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/com/google/template/soy/2019-07-14/soy-2019-07-14.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/com/google/template/soy/2019-07-14/soy-2019-07-14-jssrc_js.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/com/google/truth/truth/0.45/truth-0.45.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/org/mockito/mockito-core/3.0.0/mockito-core-3.0.0.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/com/google/javascript/closure-compiler-unshaded/v20190819/closure-compiler-unshaded-v20190819.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/net/bytebuddy/byte-buddy-agent/1.9.14/byte-buddy-agent-1.9.14.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
WARNING: Download from https://mirror.bazel.build/repo1.maven.org/maven2/org/objenesis/objenesis/3.0.1/objenesis-3.0.1.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found
[...]

PR #443, opened 2019-11-19
Mirror all http_archives in third_party_repositories.bzl.

philwo:
Context: https://github.com/bazelbuild/bazel/issues/10270


gkdn:
@alexeagle

Yannic:
@philwo Thanks! There are also some archives that have mirror urls but never got uploaded, so they probably could fail for the same reason (see #419). Could you take care of uploading these as well?

PR #444, opened 2019-11-24
Validate and convert the dependency_mode attribute.

tjgq:
Restrict the set of allowed values for js_binary.dependency_mode and convert the deprecated STRICT/LOOSE into their preferred equivalents PRUNE/PRUNE_LEGACY, so we can eventually delete the former from Closure Compiler.


tjgq:
Added a deprecation message.

PR #446, opened 2019-12-05
closure_templates_plugin: specify cfg = host

robfig:
The plugin runs on the host, and not specifying this causes cross-compilation to
fail.


robfig:
I found this when using the new rule in our codebase and attempting to cross-compile for deployment. I don't know much about platforms, but this tweak seemed to fix things.

robfig:
TIL - Updated, thanks

Yannic:
@gkdn Can you help merge this?

PR #400, opened 2019-07-05
closure_js_binary: support multiline output_wrappers via --output_wrapper_file

robfig:
Fixes #399

I believe it's not possible without using the --output_wrapper_file parameter and using that is difficult without adding support to this library itself (it can't be done in a macro, and composing rules is hard within a rule).

Ref
https://groups.google.com/forum/#!searchin/bazel-discuss/multiline%7Csort:date/bazel-discuss/GaLrGWQe-yc/XC6fQ6XWAgAJ

PR #447, opened 2019-12-10
closure_js_binary: support multiline output_wrappers via --output_wrapper_file

### LINKED ISSUE #399: closure_js_binary: output_wrapper for adding sourceMappingURL
With plovr, we had been using output_wrapper to add the sourcemap URL to compiled binaries:

```
  "output-wrapper": "(function(){%output%})();\n//# sourceMappingURL=/sourcemap?id=users_dev\n"
```

When I try the same thing with `closure_js_binary`:
```
closure_js_binary(
    name = "users_dev",
    compilation_level = "SIMPLE",
    language = "ECMASCRIPT5",
    output_wrapper = "(function(){%output%})();\n//# sourceMappingURL=/sourcemap?id=users",
```

I get an error: 
```
java.lang.IllegalArgumentException: a source must have a name
```

I believe this is due because the output_wrapper is not properly quoted in the command line that gets built up. If I try to quote it myself by adding a single quote to the beginning and end:
```
    output_wrapper = "'(function(){%output%})();\n//# sourceMappingURL=/sourcemap?id=users'",
```

I get a different error:
```
java.lang.RuntimeException: Cannot read code of a source map's source file.
Caused by: java.nio.file.NoSuchFileException: /# sourceMappingURL=/s/users/public/js/users.js.map'
```

If I escape the \n, then I end up with a literal \n in the JS file.

Should I look at submitting a pull request + fix for this? How do others add the sourcemap URL to the end of the file?

robfig:
Fixes #399

Replaces https://github.com/bazelbuild/rules_closure/pull/400. See previous discussion there.


robfig:
Not in particular -- I figured "no change" for existing users that are happy with the behavior was nice, and one less file written to disk. But I agree that it probably doesn't outweigh two paths for the same functionality.

robfig:
Updated

closure/compiler/closure_js_binary.bzl
```diff
@@ -186,9 +186,13 @@ def _impl(ctx):
         args.append("--use_types_for_optimization")
 
     if ctx.attr.output_wrapper:
-        args.append("--output_wrapper=" + ctx.attr.output_wrapper)
+        output_wrapper_file = ctx.actions.declare_file(ctx.attr.name + "_output_wrapper_file")
+        ctx.actions.write(output = output_wrapper_file, content = ctx.attr.output_wrapper)
+        inputs.append(output_wrapper_file)
+        args.append("--output_wrapper_file=" + output_wrapper_file.path)
         if ctx.attr.output_wrapper == "(function(){%output%}).call(this);":
             args.append("--assume_function_wrapper")
```
Yannic: We should probably add an attribute to `closure_js_binary` for this rather than comparing the output wrapper to an exact string, but I consider this out of scope for this PR.

robfig: Yeah, I found that strange. I don't think a new attribute is necessary though -- you can already pass it in `defs` (which is how we do it for our custom output_wrappers)

Yannic:
@gkdn Can you help to merge this?

PR #453, opened 2019-12-20
phantomjs_test: resilience improvements

robfig:
phantomjs tests fail spuriously on our CI infrastructure with two different
errors:

1. G_testRunner is not defined.

   I believe this is due to the harness setTimeout of 200ms firing before the js
   binary under test finished loading.

2. The test completes successfully, but Bazel does not get the message and
   eventually reports it as TIMEOUT.

   I believe this is due to the test runner resource fetch timing out.

These errors are addressed as follows:

1. Add an attribute to enable phantomjs debug logging.

2. Load the test runner as the last script instead of the first. That eliminates
   the race to solve issue 1.

3. Detect and log resource timeouts and retry the test. Assuming the resource
   timeouts happen with a low enough frequency, this should get tests that
   suffer one resource timeout to succeed on retry. Additionally, the logging
   will make the situation clear -- previously nothing was logged.

   If resource timeouts continue to be an issue, the next step would be to
   increase the 2s hardcoded timeout, or make it configurable.


robfig:
Hi @alexeagle ,

I think a community-maintained fork could work. I think it would still have to be coordinated with the upstream projects though, so I'm not sure it would help the situation much from your point of view.  cc @Yannic 

Swapping phantomjs for something else seems ok -- I'm not familiar enough with the area to say with what, though. I do think that having a closure_js_test rule makes sense from a scope perspective, so I would not support removing it entirely. The closure unit test framework is not a moving target and making folks develop their own runner for it seems unnecessary.  As an aside, internally some of our tests use closure unit tests and others use Jest via rules_nodejs.

Thanks

closure/testing/phantomjs_harness.js
```diff
@@ -189,6 +189,14 @@ function onLoadFinished(status) {
   }
 }
 
+/**
+ * Callback when a script fails to load in time.
+ * @param {!phantomjs.Server.ResourceError} resource
+ */
+function onResourceTimeout(resource) {
+    system.stderr.writeLine('Resource timed out: ' + resource.url);
+    retry();
```
Yannic: We probably should use Bazels flaky/retry mechanism rather than silently retrying tests, but I see that we're doing it already in some case(s), so I'm fine with doing it here as well.

PR #457, opened 2020-01-22
Upgrade Closure compiler to v20200112

gkdn:


gkdn:
@alexeagle upgrading to latest closure compiler release.

PR #459, opened 2020-01-23
Remove storage/mechanism:mechanismtestdefinition that will no longer be available

gkdn:


gkdn:
ok this is actually not needed. reverting.

PR #466, opened 2020-01-31
Fix unfurl to add an extra level of indirection

shicks:
As currently implemented, rules_closure's "strict deps checking" is inaccurate.  It recurses one level of exports and only ensures that the dep is included by something somewhere.  Recent refactoring has added an extra level of indirection to Closure Library internally, which causes this check to fail without the extra layer.  Starlark does not allow recursive function calls, so the quickest workaround is to add one more level of unfurling.  A better long-term fix would be to use a proper depset here.


shicks:
Found a better fix.

PR #468, opened 2020-02-01
Avoid coarse grained targets which will be removed in the future

gkdn:


gkdn:
I only changed our tests. I didn't remove any targets externally.
(I think we should remove the capture all targets eventually but that's different discussion)

Issue #412, opened 2019-09-04
rules_closure creates a maintenance burden for Closure Library authors


concavelenz:
Build files like https://github.com/bazelbuild/rules_closure/blob/master/closure/library/ui/BUILD are being maintained separately from Closure Library and make it possible for Closure Library maintainers to break rules_closure simply by introducing a new dependency.

gramic:
It is also a burden to the users like me, who want to keep up to date with the latest Closure Tools. However, rules_closure helps much more then it burdens me.

Maybe I can help if I get to watch closely Closure Library changes and update the rules with pull requests in faster rate.

Yannic:
I raised this issue [over a year ago](https://groups.google.com/forum/#!topic/closure-rules-discuss/8bwlPD8G6wY), but here we still are.

@concavelenz Would the Closure Library team be open to adding BUILD files in their repository? I can contribute them if necessary.

Yannic:
(For reference: google/closure-compiler#3461 google/closure-templates#190)

sgammon:
I believe @robfig has added a way to generate these files with Gazelle...

concavelenz:
@Yannic yes they are, it is just a question of getting the time to get the plumbing setup to maintain them.

robfig:
You can try out my Gazelle fork.
https://github.com/robfig/bazel-gazelle/tree/js

Check that out (the `js` branch), then `bazel build cmd/gazelle:gazelle_js` and run that from the root of your repo. 

I mean to transition it to a standalone module instead of a fork, but I'm not sure when I'm going to get to that.

alexeagle:
@robfig that looks great, I am planning to pick up this task and make these BUILD files part of the closure-library distribution. I'll try out your gazelle plugin.

alexeagle:
@robfig I tried that out yesterday, totally works. We are thinking about a couple things though:
- it would be good if closure-library can be consumed without rules_closure since the latter is kinda unmaintained, compared with rules_nodejs that supports the entire JS ecosystem. So I think we should use the `js_library` rule from rules_nodejs - I need to promote it to the public API first
- We have BUILD files internal at Google for closure-library. @nreid260 points out that these are the only ones which are really tested, and the naming convention of files is not 100% related to the goog.module names. So maybe we should just scrub/publish our BUILD files with closure-library releases

robfig:
[...]

PR #475, opened 2020-03-03
Generate BUILD files for closure-library during loading/analysis phase

Yannic:
This change introduces a new repository rule to generate the BUILD files
for closure-library when the repository is downloaded instead of relying
on checked-in versions (i.e. the files we have in `//closure/library`).

This will eventually require that users migrate their WORKSPACE to
reference the targets directly from the closure library instead of the
ones from this repo. We will provide a buildozer command in the release
notes that can be used to (mostly) do the migration. In the meantime,
we'll update the targets in `//closure/library` to point to the new
targets.

Updates #412


gkdn:
Hi Yannic. I didn't understand the value of this intermediate step?

gkdn:
Closing since this is obsolete now.

Issue #421, opened 2019-09-14
Make rules_closure conform to Style Guide for Rules


Yannic:
Tracking bug for making rules_closure conform to Bazel's [Style Guide for Rules](https://docs.bazel.build/versions/master/skylark/deploying.html).

Known issues:
  - [ ] Dependencies / Toolchains should be declared by `rules_closure_{dependencies,toolchains}` in `closure/repositories.bzl`: https://docs.bazel.build/versions/master/skylark/deploying.html#dependencies

Yannic:
@alexeagle Can you handle the repo description issue?

alexeagle:
Sure, done, thanks for picking this up.

Yannic:
Thank you!

PR #476, opened 2020-03-03
Remove deprecated closure_repositories() macro

Yannic:
This was deprecated in favor of `rules_closure_dependencies()` and
`rules_closure_toolchains()` to be more consistent with other rules.
See https://docs.bazel.build/versions/master/skylark/deploying.html#dependencies

Updates #421

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,9 +90,6 @@ rules_closure_toolchains()
 You are not required to install the Closure Tools, PhantomJS, or anything else
 for that matter; they will be fetched automatically by Bazel.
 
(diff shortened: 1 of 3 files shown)


Yannic:
I did remove the warning that 0.10 is the last release to support `closure_repositories()` and couldn't resist replacing `python` with `starlark`.

gkdn:
Do you mind separating those two patches?

Yannic:
Thanks, done! Split `python -> starlark` change into https://github.com/bazelbuild/rules_closure/pull/479

PR #487, opened 2020-06-17
Update closure compiler version to v20200614

jDramaix:


jDramaix:
Some test are failing. I'm investigating

PR #491, opened 2020-06-23
Stop leveling all diagnostics as ERROR by default

nreid260:
Updates the leveling algorithm in JsCompilerWarnings to ignore any diagnostics it doesn't recognize, rather than assuming they're errors. Other guards are now given the chance to level such diagnostics. If there are no other guards, the the intrinsic level of the diagnostic is used.

This makes it possible to roll out new JSCompiler diagnostics without automatically breaking rules_closure users.

### DIFF
--- a/java/com/google/javascript/jscomp/JsCompilerWarnings.java
+++ b/java/com/google/javascript/jscomp/JsCompilerWarnings.java
@@ -57,8 +57,6 @@ final class JsCompilerWarnings extends WarningsGuard {
 
   @Override
   public CheckLevel level(JSError error) {
-    CheckLevel level = CheckLevel.ERROR;
-
     // Closure Rules will always ignore these checks no matter what.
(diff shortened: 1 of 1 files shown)


nreid260:
I've updated the PR to simply respect OFF diagnostics. I think this gives us the flexibility we want with out the confusion or regression of the first version.

gkdn:
👍

pcj:
Finally upgrading past this commit and quite surprised by this.  rules_closure since the beginning @jart has always had a strict-by-default policy, I really feel this change goes against the culture of this project.  Can we revert this please?

How do I re-enable all the checks I had before?  Having trouble trusting the compiler output now.

gkdn:
This had too much unnecessary burden to Closure Compiler developers who needed to update this project every time they introduce a new diagnosis, that was an unreasonable ask from them.
If that issue is addressed in the future (eg. the list lives in the compiler), we can consider going back to previous state.

PR #494, opened 2020-06-27
Remove (deprecated) implicit outputs, part 1

Yannic:
The `outputs` parameter on the `rule()` function has been deprecated and
will be removed in a future version of Bazel.

See https://github.com/bazelbuild/bazel/issues/7977 for context.


gkdn:
Taking a quick look at linked bug, it seems like it is still possible to preserve labeled output - otherwise it would be impractical to migrate for many.
Unless I'm missing something; preserving label output will help backward compatibility and reduce amount of changes in tests. Any reason not to do that?

gkdn:
Thanks.

PR #507, opened 2021-01-14
Remove last remaining reference to removed Closure Compiler diagnostic group.

blickly:
The CheckMissingAndExtraRequires pass that this diagnostic referred to has been removed in favor of the
CheckMissingRequires pass, which can now be turned on via the more simply named "missingRequire" diagnostic group.


gkdn (approved):
This will probably also require compiler version upgrade but that could be done later.

blickly:
Thanks Goktug!

PR #517, opened 2021-04-21
Filter out non-JavaScript files when expanding tree artifacts.

tjgq:


tjgq:
@gkdn can you review this?

The code looks a little weird because the `get_jsfile_path` helper is also used in Google-internal code (that I'm going to touch next). Once J2CL has been switched to tree artifacts and we no longer need to support zip files, we can use `get_jsfile_path` instead of the `_get_src_path` helper in `closure_js_binary.bzl`.

gkdn (approved):

PR #519, opened 2021-04-30
Add user flags before srcs file on compiler command line.

jDramaix:
User can use the rule argument `defs` in order to pass specific flags to the closure compiler. Those flags need to be passed on the command line before the list of srcs file otherwise they are considered a srcs file name and compilation fails.


gkdn (commented):
Looks good but could you add a test?

jDramaix:
> Looks good but could you add a test?

Done

jDramaix:
I'm closing this issue. I was able to reproduce my issue but this is not solved with this PR. Options/arguments order is not an issue with args4j.

PR #520, opened 2021-05-04
Remove isPrivate reference in advance of its removal

concavelenz:
isPrivate is being remove from Closure Compiler


gkdn:
I think you will need to keep it but remove the override for now.

gkdn (approved):

PR #524, opened 2021-05-26
Eliminate deps on HotSwapCompilerPass

nreid260:
HotSwapCompilerPass and the concept of hostswapping are being removed from Closure Compiler.


jDramaix (approved):
Could you just add in the description why this change is needed ?

nreid260:
Amended description.

gkdn (approved):

PR #537, opened 2021-10-19
Simplify closure_proto_library include path generation

aarossig:
Use built-in facilities for generating transitive import paths. This resolves a bug with handling generated protos (specifically: `any.proto` from the protobuf project). Without this fix, the rebuilt proto import paths are incorrect and result in errors.

Example:

```
google/protobuf/any.proto: File not found.
google/rpc/status.proto:19:1: Import "google/protobuf/any.proto" was not found or had errors.
google/rpc/status.proto:46:12: "google.protobuf.Any" is not defined.
```

The issue is caused by the fact that `any.proto` is generated and the path for it is virtual. The logic in the previous implementation is unable to rebuild a path like this.

```
<generated file external/com_google_protobuf/_virtual_imports/any_proto/google/protobuf/any.proto>
```

### DIFF
--- a/closure/protobuf/test/BUILD
+++ b/closure/protobuf/test/BUILD
@@ -98,3 +98,39 @@ closure_js_test(
         "//closure/library/testing:testsuite",
     ],
 )
+
+proto_library(
+    name = "bat_proto",
+    srcs = ["bat.proto"],
+    deps = ["@com_google_protobuf//:any_proto"],
+)
+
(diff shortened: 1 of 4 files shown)


aarossig:
All set. The test is fairly simple. The proto simply won't compile with a dependency on any proto. I cherry-picked the test patch without the fix and the failure is the same that I see in my own project.

PR #540, opened 2021-11-05
Explicitly set the `--emit_use_strict` flag

blickly:
This makes this test case compatible with an upcoming compiler change that makes `--emit_use_strict` false by default.


gkdn:
Looks like the tests is still failing. Did you take a look at the failure?

PR #541, opened 2021-11-09
Updates Closure Compiler & Library to latest versions.

gkdn:
This also updates protobuf and Closure template dependencies.

The patch removes //closure/library package since Closure Library recently
updated to include necessary BUILD targets under @com_google_javascript_closure_library//closure/goog.

For migration:
Most users can do replace @io_bazel_rules_closure//closure/library with
@com_google_javascript_closure_library//closure/goog.
Note that :all_js targets are not provided and can be replaced with specific
targets in the particular package.

Co-authored-by: pcj

PR #542, opened 2021-11-10
Explicitly add --emit_use_strict

lauraharker:
This is a no-op with older Closure Compiler releases because of a command line runner bug, where `--emit_use_strict=true` is ignored. That's why https://github.com/bazelbuild/rules_closure/pull/540
had test failures. (Also, @blickly  is on leave so can't update that PR)

This PR fixes the Google-internal CI which runs the compiler at head.
Keeping `language = ECMASCRIPT5_STRICT` avoids the CI failures from https://github.com/bazelbuild/rules_closure/pull/540, but is unnecessary with the latest Closure Compiler version.

Landing https://github.com/bazelbuild/rules_closure/pull/541 would make
this PR unnecessary. But I'd like to submit this earlier if possible to fix the Google CI.

### DIFF
--- a/closure/compiler/test/BUILD
+++ b/closure/compiler/test/BUILD
@@ -38,8 +38,8 @@ file_test(
 
 file_test(
     name = "sourcemap_doesntContainWeirdBazelDirectories",
-    regexp = r'"sources":\["/closure/compiler/test/hello.js"\]',
     file = "hello_bin.js.map",
+    regexp = r'"sources":\["/closure/compiler/test/hello.js"\]',
 )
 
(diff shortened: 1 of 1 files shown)


gkdn:
@lauraharker I have a proper fix for this but I need the new Closure compiler release. Do you happen to know why there was a delay in the release?

lauraharker:
That's embarrassing... I was supposed to do the Closure Compiler release
this month, based on the 2021-11-03 build, but completely forgot.

I'll start the release today.

lauraharker:
Ok, Closure Compiler v20211107 should be available.

PR #558, opened 2022-06-08
Add attribute to web_library for host:port

dsyer:
Example (from test):

```
web_library(
    name = "raven",
    srcs = ["index.html"],
    path = "/",
    deps = [":stuff"],
    host = "0.0.0.0",
    port = "8080"
)
```

(Supersedes #557.)

### DIFF
--- a/closure/webfiles/web_library.bzl
+++ b/closure/webfiles/web_library.bzl
@@ -128,7 +128,7 @@ def _web_library(ctx):
     # define development web server that only applies to this transitive closure
     params = struct(
         label = str(ctx.label),
-        bind = "[::]:6006",
+        bind = str(ctx.attr.bind),
         manifest = [long_path(ctx, man) for man in manifests.to_list()],
         external_asset = [
             struct(webpath = k, path = v)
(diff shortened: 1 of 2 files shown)


closure/webfiles/web_library.bzl
```diff
@@ -209,6 +209,7 @@ web_library = rule(
     executable = True,
     attrs = {
         "path": attr.string(),
+        "bind": attr.string(default = "0.0.0.0:6006"),
```
gkdn, line 212: Let's introduce `host` and `port` attributes; I think it will be more self documenting and in many cases users will only change the port or the host.

gkdn (approved):

closure/webfiles/web_library.bzl
```diff
@@ -209,6 +209,8 @@ web_library = rule(
     executable = True,
     attrs = {
         "path": attr.string(),
+        "host": attr.string(default = "0.0.0.0"),
+        "port": attr.string(default = "6006"),
```
gkdn, line 213: I just noticed that this is `attr.string` instead of `attr.int` which requires usages to pass `"8080"` instead of `8080`.

PR #559, opened 2022-07-13
Remove calls to deprecated `setFeatureSet(FeatureSet)` which does not…

rishipal:
…hing and will be deleted


soldair (approved):
lgtm

rishipal:
This PR causes this breakage. It's happening because the internal PassFactory.java contains the `setFeatureSet` in its builder class.

```
java.lang.IllegalStateException: Missing required properties: featureSet
--
  | at com.google.javascript.jscomp.AutoValue_PassFactory$Builder.autoBuild(AutoValue_PassFactory.java:150)
```
I think I should land the internal change which deletes the `setFeatureSet` method first, and then immediately fix the upstream. What do you think?

gkdn:
@rishipal You can land the internal change. There is another similar change that was submitted internally. 

I will fix them together in the opensource with the next update.

gkdn:
Submitted in https://github.com/bazelbuild/rules_closure/commit/6bbbfc7d80474a6a1a4491da7b601c0cab440b07

PR #588, opened 2023-09-25
Supply data to WebfilesServer in webdriver_test.

mollyibot:


closure/testing/webdriver_test.bzl
```diff
@@ -45,18 +46,31 @@ def webdriver_test(
...
+        path = path,
+        testonly = True,
+    )
+
     # set up a development web server that links to the test for debugging purposes.
     web_library(
         name = "%s_debug" % name,
-        srcs = [html, test_file_js],
+        srcs = data if data else [],
+        deps = [":%s_test_files" % name],
```
gkdn, line 60: Can we just use exports here instead of changing the web_library rule?

mollyibot: I think even exports we still need deps, I do not see exports is handled for webpath https://source.corp.google.com/piper///depot/google3/third_party/bazel_rules/rules_closure/closure/webfiles/web_library.bzl;l=47?q=web_library

gkdn (approved):

PR #591, opened 2023-10-06
Modify test driver to wait until test finishes.

mollyibot:
Wasm junit test running on mac system takes a bit long. the test has not finished yet when test driver detect the finish status, but it runs successfully in the end.

### DIFF
--- a/java/io/bazel/rules/closure/testing/TestDriver.java
+++ b/java/io/bazel/rules/closure/testing/TestDriver.java
@@ -56,6 +56,7 @@ public class TestDriver {
           );
     } catch (TimeoutException e) {
       logger.log(Level.SEVERE, String.format("Test timeout after %s seconds", TEST_TIMEOUT));
+      return false;
     }
 
     String testReport =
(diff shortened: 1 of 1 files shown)


java/io/bazel/rules/closure/testing/TestDriver.java
```diff
@@ -53,23 +52,12 @@ public boolean run() {
...
-                if (!testFinishedSuccessfully) {
-                  logger.log(Level.SEVERE, "G_testRunner has not finished successfully");
-                }
-                return true;
-              });
+              executor -> (boolean) executor.executeScript("return window.top.G_testRunner.isFinished()")
+          );
     } catch (TimeoutException e) {
-      testFinishedSuccessfully = false;
       logger.log(Level.SEVERE, String.format("Test timeout after %s seconds", TEST_TIMEOUT));
```
gkdn, line 58: We also need to return false here. There is no much point of trying to dig the results if it has timed out

mollyibot: returned false if time out.

gkdn (approved):

PR #596, opened 2023-11-01
Disable bzlmod for rules_closure as bazel now enables bzlmod by defaul…

mollyibot:
…t and some tests fail

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,1 @@
+build --noenable_bzlmod


.bazelrc
```diff
@@ -0,0 +1 @@
+build --noenable_bzlmod
```
meteorcloudy, line 1: Change to `common` or `always`?

meteorcloudy (changes requested):

PR #599, opened 2024-01-17
Replace legacy struct providers with modern ones

comius:
Legacy struct providers have been deprecated by Bazel. Replacing them with modern providers, will make it possible to simplify and remove legacy handling from Bazel


closure/private/defs.bzl
```diff
@@ -61,6 +61,84 @@ CLOSURE_JS_TOOLCHAIN_ATTRS = {
...
+NestedSet<Label> of all closure_css_library rules in the transitive
+closure. This is used by closure_js_binary can guarantee the
+completeness of goog.getCssName() substitutions.""",
+    "has_closure_library": """
+Boolean indicating indicating if Closure Library's base.js is part
+of the srcs subprovider. This field exists for optimization.""",
+    "language": "",
+})
+
+ClosureExportsInfo = provider("ClosureExportsInfo", fields = {"exports": """
```
gkdn, line 110: Since we are introducing proper providers, would it be possible to get rid of this as well?

comius: For `ClosureExportsInfo`, it looks possible to remove it, by adding `exports` to `ClosuresJsLibraryInfo` and `WebFilesInfo`.

`ClosureJsLegacyRunfilesInfo` looks easier to remove, by using regular runfiles instead.

Both of those removals require some effort and should probably be done by code owners and not as a part of a large cleanup, where we're fixing about 80 other rulesets.

gkdn: Ok that's fair.

gkdn:
@mollyibot Could you do a quick/hacky prototype of required changes to [build_defs/internal_do_not_use/j2cl_js_common.bzl ](https://github.com/google/j2cl/blob/master/build_defs/internal_do_not_use/j2cl_js_common.bzl) to validate this.

comius:
> I did repro build for j2cl/ j2wasm, One minor thing : some deps may still in deprecated struct which means the[ in operator ](https://bazel.build/rules/lib/builtins/Target#modules.Target.providers) or index notation would not always work.

We have internal fixes already for that file. They seem independent of rules_closure.

gkdn:
> > I did repro build for j2cl/ j2wasm, One minor thing : some deps may still in deprecated struct which means the[ in operator ](https://bazel.build/rules/lib/builtins/Target#modules.Target.providers) or index notation would not always work.
> 
> We have internal fixes already for that file. They seem independent of rules_closure.

To double check, you are referring to fixes in Bazel, right? That means if we submit the change we cannot use it from older versions?

comius:
> > We have internal fixes already for that file. They seem independent of rules_closure.
> 
> To double check, you are referring to fixes in Bazel, right? That means if we submit the change we cannot use it from older versions?
> 
> Edit: more particularly J2CL that is using `5.4.0`?

No, not Bazel. The support for new style providers is in Bazel for ages, I believe even before Bazel 4.

I was referring to fixes of `j2cl_js_common.bzl`, they are done internally.

gkdn:
@mollyibot What is the source of non-target deps here?

gkdn:
But that's because we haven't migrated J2CL, right? i.e. j2cl is still producing the struct

gkdn:
But you mentioned that you needed to workaround a problem on J2CL side, I thought it was something else. Could you share me again the changes needed in J2CL?

> for other callers that are not from j2cl

We gonna need to be move forward at one point. Maybe we can temporary support both but one can just wait before upgrading since we don't have other important improvement. Anyway we can discuss if that becomes the only problem (your PR is small so it might be a good tradeoff)

gkdn:
IIUC your concern is following steps:
1. rules_closure will be changed with this PR.
2. We need to update J2CL to work with both old version and new version (legacy struct and provider) - i.e. the messy workaround
3. J2CL open-source will be updated to new rules_closure version
4. J2CL will be updated again to remove the workaround

And you are proposing:
1. rules_closure changed to support both versions - legacy struct and provider  (this PR + your PR)
2. J2CL is updated to work with providers
3. rules_closure support for legacy struct is dropped

And the final state in both approaches will be the same.

Is that accurate or do you have other concerns?

gkdn:
There shouldn't be any struct passing scenario; we are under full control and that would be the point of the migration. That also shouldn't happen under js_checkable_provider either. I think something might be wrong in this picture if that's not the case.

gkdn:
Ok plan wise; 
let's do a release of rules_closure and point out that this will be the last version to support existing legacy style. I can do this, when I get back from vacation.

Then we can submit this PR and do follow up changes to do further refactoring to get rid of the extra providers discussed in https://github.com/bazelbuild/rules_closure/pull/599#discussion_r1456810215

Then prepare the J2CL PR that only supports the latest version of rules_closure. If there is any surprising impact; we can do more updates to rules_closure and finalize the next steps.

Does it sound good?

gkdn (approved):

PR #601, opened 2024-02-29
Replace legacy struct providers with modern ones

comius:
Legacy struct providers have been deprecated by Bazel.


comius:
Hey @gkdn, I forgot, I've amended the internal change with this part as well.

PR #604, opened 2024-03-06
Make closure_js_library to take deps in provider on top of target

mollyibot:
I take some efforts to refactor the code to take only provider, in that approach, we need to change for all providers and `collect_runfiles` and `data` , this kind of break the concept of https://github.com/bazelbuild/rules_closure/pull/599 that pass target everywhere. I think for now this small patch should help us to walk around the issue that closure_js_library only handles target(but j2cl pass provider).

### DIFF
--- a/closure/compiler/test/linting/BUILD
+++ b/closure/compiler/test/linting/BUILD
@@ -21,7 +21,10 @@ load("//closure:defs.bzl", "closure_js_library")
 
 closure_js_library(
     name = "love",
-    srcs = ["love.js"],
+    srcs = [
+        "love.js",
+        "somemodule.js",
+    ],
     internal_expect_failure = True,
 )
 
@@ -29,6 +32,12 @@ file_test(
     name = "methodMissingJsdoc_causesError",
     file = "love-stderr.txt",
     regexp = "JSC_MISSING_JSDOC",
+)
+
+file_test(
+    name = "checksGoogModule",
+    file = "love-stderr.txt",
+    regexp = "somemodule.js",
 )
 
 closure_js_library(
@@ -47,7 +56,10 @@ file_test(
 
 closure_js_library(
     name = "lenient_love",
-    srcs = ["love.js"],
+    srcs = [
+        "love.js",
+        "somemodule.js",
+    ],
     lenient = True,
 )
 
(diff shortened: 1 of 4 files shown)


gkdn:
could you do
```
dep = dep[ClosureJsLibraryInfo] if type(dep) == "Target" else dep
```
try to keep rest of the code same.

closure/private/defs.bzl
```diff
@@ -135,12 +135,20 @@ def unfurl(deps, provider = ""):
...
-        if not provider or provider in dep:
-            res.append(dep)
-        if type(provider) == "Provider" and provider in dep and hasattr(dep[provider], "exports"):
-            for edep in dep[provider].exports:
-                if not provider or provider in edep:
-                    res.append(edep)
+        if type(dep) == "Target":
+            if not provider or provider in dep:
+                res.append(dep)
+            if type(provider) == "Provider" and provider in dep and hasattr(dep[provider], "exports"):
```
gkdn, line 141: do we ever pass non provider as a provider here?

gkdn:
> > could you do
> > ```
> > dep = dep[ClosureJsLibraryInfo] if type(dep) == "Target" else dep
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > try to keep rest of the code same.
> 
> for other providers we need to keep target format, for example https://github.com/comius/rules_closure/blob/daf6131a64793e4f684fb9658ce7b6e5726896fb/closure/compiler/closure_js_library.bzl#L168-L182
> 
> ```
> deps = unfurl(deps, provider = ClosureJsLibraryInfo).exports
> 
>     # Collect all the transitive stuff the child rules have propagated. Bazel has
>     # a special nested set data structure that makes this efficient.
> 	@@ -177,7 +179,7 @@ def _closure_js_library_impl(
>     # which is a superset of the CSS libraries in its transitive closure.
>     stylesheets = []
>     for dep in deps:
>         if ClosureCssLibraryInfo in dep:
>             stylesheets.append(dep.label)
> ```
> 
> we can do this for closureJsProvider related though.

Can't we do unfurl them in the same way:
```
    cssDeps = unfurl(deps, provider = ClosureCssLibraryInfo)
    for cssDep in cssDeps:....
```

closure/private/defs.bzl
mollyibot: yes, the default parameter is "", the function signature is 
```
def unfurl(deps, provider = ""):
```
sometimes there is only one parameter without provider.

gkdn:
The unfurl is overloaded with multiple purposes and that makes a lot of things obscure. I did a refactoring around that, will send for review but first I sent you a cleanup for runfiles handling.

PR #607, opened 2024-03-07
Make create_closure_js_library work on providers.

gkdn:
This follows a more common provider approach where rules are
responsible for providing supplier to helper functions vs. helper
functions trying make it workaround with a mix of targets and
providers.

This also unbundles provider extraction from unfurl to remove
implicit and optional behaviors behind it.

### DIFF
--- a/closure/compiler/closure_js_binary.bzl
+++ b/closure/compiler/closure_js_binary.bzl
@@ -30,6 +30,7 @@ load(
     "get_jsfile_path",
     "sort_roots",
     "unfurl",
+    "extract_providers",
 )
 load(
     "//closure/compiler:closure_js_aspect.bzl",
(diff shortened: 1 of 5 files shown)


gkdn:
@mollyibot

closure/compiler/closure_js_binary.bzl
```diff
@@ -30,6 +30,7 @@ load(
     "get_jsfile_path",
     "sort_roots",
     "unfurl",
+    "extract_providers",
```
mollyibot, line 33: [nit] order above unfurl?

gkdn: Good catch. Ran buildifier on edited files.

jDramaix (approved):

PR #618, opened 2024-09-23
Add `load()`s for the java rules

hvadehra:


gkdn (approved):

hvadehra:
@gkdn Is there a periodic import or is that something I need to do internally?

gkdn:
Taking care of it.

PR #619, opened 2024-09-27
Upgrade closure_library to v20230802

mollyibot:


closure/BUILD
```diff
@@ -45,3 +45,5 @@ bzl_library(
         "//closure/testing:__pkg__",
     ],
 )
+
+exports_files(["soy_build.patch"])
```
gkdn, line 49: are you sure we need this?

closure/templates/BUILD
```diff
@@ -33,6 +33,7 @@ closure_js_library(
     name = "soy_jssrc",
     srcs = ["@com_google_template_soy_jssrc"],
     lenient = True,
+    suppress = ["missingSourcesWarnings"],
```
gkdn, line 36: could you clarify what this is?

closure/BUILD
mollyibot: not neccessary. Removed.

closure/templates/BUILD
mollyibot: this is coming from a type check error:
```
external/com_google_template_soy_jssrc/soyutils_usegoog.js:149: ERROR - Property getDirection never defined on SafeHtml
        goog.html.SafeHtml.unwrap(value), value.getDirection());
                                                ^
  ProTip: "JSC_INEXISTENT_PROPERTY" or "checkTypes" or "missingProperties" or "missingSourcesWarnings" can be added to the `suppress` attribute of:
  @@//closure/templates:soy_jssrc
  Alternatively /** @suppress {missingSourcesWarnings} */ can be added to the source file.

```

gkdn: This is not something we can safely ignore. Could you try upgrade soy to `36c7400e51c71c0af2b8b569e613a15f05deb994`?
If doesn't work, then you can try `309707311325543e936fd14df24dd026c3051737`.

mollyibot: I tried both and neither works, and both give error :
```
xternal/com_google_template_soy_jssrc/soydata_converters.js:47: ERROR - Namespace not provided by any srcs or direct deps of @@//closure/templates:soy_jssrc.
const {ByteString} = goog.require('jspb.bytestring');
                                  ^
  ProTip: "CR_NOT_PROVIDED" or "strictDependencies" can be added to the `suppress` attribute of:
  @@//closure/templates:soy_jssrc
```

mollyibot: I also tried other newer version but so far have not found a candidate yet.

gkdn: Ok jspb dependency introduced then the directionality is removed so there wouldn't be a working version.

Can try to patch in [this diff](https://github.com/google/closure-templates/commit/309707311325543e936fd14df24dd026c3051737) in http_archive to see if it is enough?

mollyibot: I patch this diff in the repositiory PTAL, but I feel it does not bring us more gain between just suppresssing it vs maintaining the patch

gkdn: You are not simply suppressing a warning; it is invalid call that accessing a property that doesn't exist. The patch is fixing that.

Regardless we will remove soy support after the release. I just wanted to make sure we don't make a release with soy completely broken.

gkdn (approved):

closure/templates/BUILD
mollyibot: Make sense! thanks for the clarification.

PR #620, opened 2024-10-24
Upgrade JsCompiler Version to v20240317

mollyibot:


gkdn:
No objections if it doesn't break anything in rest of the toolchain. Have you checked that?

gkdn:
With toolchain I'm referring to other tools we have like jsinterop genertor and elemental2

jDramaix:
if you can build elemental2 with this patch, it's LGTM

PR #623, opened 2024-11-13
Use the Bazel zip tool instead of the jar tool.

gkdn:
The jar tool is deprecated.


gkdn:
@mollyibot or @jDramaix  could you approve?

mollyibot (approved):

PR #633, opened 2024-12-04
Remove goog.string dependency.

gkdn:


closure/compiler/test/goog_es6_interop/string.js
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * @fileoverview A goog.provide file for testing interoperability.
+ */
+goog.provide('person.string');
+
+/**
+ * @param {string} str String value to capitalize.
+ * @return {string} String value with first letter in uppercase.
+ */
+person.string.capitalize = function(str) {
```
jDramaix, line 24: Could we simply add this method method to person.js instead?

gkdn: The test is for testing imports across es6 modules and goog.provide.

jDramaix: Ah I missed that point =)

jDramaix (approved):

mollyibot (approved):

PR #639, opened 2025-01-08
Remove protobuf support.

gkdn:
With archiving of Closure Libraries and official implementation
still relying on them makes it impractial to continue supporting
protos at the moment. For now, I will remove the support so we
can move forward with migration but this will be something we
need to figure this out later on.


README.md
```diff
@@ -2,9 +2,9 @@
 
 JavaScript | Stylesheets | Miscellaneous
 --- | --- | ---
-[closure_js_library] | [closure_css_library] | [closure_js_proto_library]
-[closure_js_binary] | [closure_css_binary] | [phantomjs_test]
-[closure_js_test] | | [closure_proto_library] \(Experimental\)
+[closure_js_library] | [closure_css_library] | [phantomjs_test]
```
jDramaix, line 5: nits: I would put `phantomjs_test` on the same line than `closure_js_test`

jDramaix (approved):

README.md
gkdn: I don't want to put meaning into rows since it is not necessarily feasible to maintain that.

PR #658, opened 2025-04-11
Update closure-compiler to v20250402

kevinoconnor7:


.bazelrc
```diff
@@ -0,0 +1,5 @@
+# Compile/run with Java 21
```
kevinoconnor7, line 1: The closure-compiler jars pushed to maven are compiled with Java 21.

closure/compiler/closure_js_library.bzl
```diff
@@ -140,7 +140,6 @@ def _closure_js_library_impl(
             "strictCheckTypes",
             "strictModuleChecks",
             "superfluousSuppress",
-            "unnecessaryEscape",
```
kevinoconnor7, line 143: This is now included in `lintChecks`

java/com/google/javascript/jscomp/JsCheckerPassConfig.java
```diff
@@ -111,13 +110,6 @@ private PassFactory scopedAliases() {
         .build();
   }
 
-  private PassFactory closureRewriteClass() {
-    return PassFactory.builder()
-        .setName("closureRewriteClass")
-        .setInternalFactory((compiler) -> new ClosureRewriteClass(compiler))
```
kevinoconnor7, line 117: `ClosureRewriteClass` has since been deleted. It was used to support `goog.defineClass`.

jDramaix (approved):

.bazelrc
```diff
@@ -0,0 +1,5 @@
...
+common --java_language_version=21
+common --java_runtime_version=remotejdk_21
+common --tool_java_language_version=21
+common --tool_java_runtime_version=remotejdk_21
```
gkdn, line 5: Let's follow https://github.com/google/j2cl/blob/master/.bazelrc

kevinoconnor7: Done

gkdn:
I am checking it again and again but I only see 1 commit and no new changes in the review, although this page suggests otherwise. I think you might have done something unusual while pushing it.

kevinoconnor7:
oops sorry. I didn't include the change I intended when I squashed. Should be good now.

gkdn:
For the next time, let's not squash while pushing since it removes the history for code review, you can't compare (merging will squash it regardless)

PR #659, opened 2025-05-21
Update rules closure to not depend on rules_scala

mollyibot:


MODULE.bazel
```diff
@@ -45,15 +45,19 @@ bazel_dep(
...
     remote = "https://github.com/bazelbuild/rules_webtesting",
 )
 
-# TODO(mollyibot): Needed by rules_webtesting.
-git_override(
-    module_name = "rules_scala",
-    commit = "219e63983e8e483e66ebf70372969ba227382001",
-    remote = "https://github.com/mbland/rules_scala",
+bazel_dep(
+    name = "rules_web_testing_java",
```
jDramaix, line 53: you should be consistent and use `rules_webtesting` prefix everywhere

mollyibot: I update this module to be rules_webtesting prefix but this is submodule web_testing_java and we cannot use prefix for git_override submodule.

mollyibot: https://github.com/bazelbuild/rules_webtesting/blob/9da96ee35b478bff4ab02795551f1a89a693e864/web_testing_java/MODULE.bazel#L2

gkdn: I think Julien is asking if we can preserve the implicit name (`rules_web_testing_java`) instead of overriding it via `repo_name`.

gkdn:
Is there any value of pushing this before uploading rules_webtesting to bcr?

(If it is for testing, I guess you can still do it locally without merging this?)

gkdn:
But is there any benefit from merging this pr? IIRC, the reason we started with dropping scala was bcr related.

PR #660, opened 2025-05-23
Register rules_closure in bcr

mollyibot:


.bcr/presubmit.yml
```diff
@@ -0,0 +1,10 @@
+bcr_test_module:
+  matrix:
+    platform: [ "macos", "ubuntu1804"]
+    tasks:
+      run_tests:
+        name: "Run test module"
```
kevinoconnor7, line 6: i'd remove the name... it's not clarifying anything

.bcr/presubmit.yml
```diff
@@ -0,0 +1,10 @@
+bcr_test_module:
+  matrix:
+    platform: [ "macos", "ubuntu1804"]
```
kevinoconnor7, line 3: this looks misconfigured. `tasks` shouldn't be nested under `matrix`

kevinoconnor7, line 3: Use a list rather than array

kevinoconnor7, line 3: why such an old ubuntu version?

.bcr/presubmit.yml
```diff
@@ -0,0 +1,10 @@
+matrix:
+  platform:
+    - macos
+    - ubuntu2004
+tasks:
+  run_tests:
+    build_targets:
```
kevinoconnor7, line 7: missing: `platform: ${{ platform }}`

Also should we be testing against bazel 7 and 8 like the other repos?

kevinoconnor7 (changes requested):

kevinoconnor7 (approved):

PR #664, opened 2025-06-13
Update bcr to include test module and add windows platform

mollyibot:
based on comments here , i added back the test module and update the windows system.


kevinoconnor7:
I don't see windows being added... I see debian

PR #665, opened 2025-09-09
Add a null check to Long.equals

kevinoconnor7:
`Long.equals` accepts a nullable parameter, but will crash if you actually pass one. Instead we should always return null if `other` is null/undefined.

This is a backport of the internal version.


gkdn (approved):
Let's clarify in PR/commit this is a sync back.

